Issue #6, opened 2019-04-29
readthedocs is improperly configured


rafmudaf:
There are a variety of errors generated when this project is linked to readtehdocs. Here are two examples: [mine](https://readthedocs.org/projects/raf-openoa/) and [openoa](https://readthedocs.org/projects/openoa/). Some of the errors are as simple as unmet requirements, while others are more subtle.

jordanperr:
Up and running now. Thanks Rafael! This configuration requires entry of "sphinx/requirements.txt" in the ReadTheDocs settings.

https://openoa.readthedocs.io/en/latest/

PR #7, opened 2019-04-29
Fix some errors in the sphinx-readthedocs configuration

rafmudaf:
This pull request fixes errors in the Sphinx configuration for readthedocs and closes issue #6.

These are the updates included here:
- Some requirements were missing from the documentaiton-specific requirements file so that is updated
- An extension to convert markdown to rst, `m2r`, has been added to the sphinx configuration file
- An invalid python lexer, `ipython2`, was changed to `python`
- The biggest fix: updating the sphinx version to 2.0.0

Note: these updates get the build to finish successfully, but there are still errors within the documentation. For example, the python lever is not appropriate for all code blocks so the appropriate ones from the [pygments documentation](http://pygments.org/docs/lexers/) should be chosen.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -101,7 +101,7 @@ python setup.py integrate -a "--junitxml=./path_to_outputfile.xml"
 Documentation is provided by [sphinx](http://www.sphinx-doc.org/en/stable/). To (re)build the documentation:
 
 ```
-pip install sphinx_rtd_theme ipython m2r
+pip install sphinx_rtd_theme ipython
 
 cd sphinx
 make html
--- a/requirements-python3.txt
+++ b/requirements-python3.txt
@@ -1,3 +1,4 @@
+# Requirements for the OpenOA software package
 numpy==1.15.4
 scipy==1.1.0
 pandas==0.23.4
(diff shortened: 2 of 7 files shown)


jordanperr:
Thanks for your contribution! This looks good to me. I downloaded and tested this to verify the documentation builds correctly. A couple questions:

1. Why did you change all of the cells to "python" from "ipython"? This was the output of a jupyter nbconvert, where cells are interpreted through an ipython kernel.
2. Are the requirements.txt and sphinx/requirements.txt meant to be composed in a development environment? That is, should developers now install from both requirements files when setting up a new development environment? Why not just include the docs dependencies in the top-level requirements.txt?

rafmudaf:
1. On my local build and on readthedocs, the lexer `ipython2` wasnt rendering and I didn't see it in this list of [lexers](http://pygments.org/docs/lexers/).

2.  I admit I may be way off, but I assume that users wont be building the documentation locally and will be going to readthedocs instead. So I left your docs specific requirements file with only the packages required to build it to minimize problems on the readthedocs build image. Actually, I reworked the FLORIS docs [requirements file](https://github.com/NREL/floris/blob/develop/docs/requirements.txt) similarly but added a path back to the main requirements file. This way, users don't have to install all of the sphinx stuff. 
The FLORIS situation is slightly different in that all of the packages in the main requirements file are required to build the docs. In OpenOA, I was able to build the docs locally and on readthedocs with only the packages in `sphinx/requirements.txt`.

Issue #10, opened 2019-04-30
Air density calculation differs from IEC 61400-12-1 / 2


sebastianpfaffel:
The air density calculation differs in two points from the IEC 61400-12-/2 definition.

- [x] The calculation should consider humidity and use a default value of 0.5 in case of missing values. By not considering humidity it is currently set to 0.
 - [x] The gas constant for dry air used by the IEC standard is 287.05 instead of 287.058. The IEC definition is less precise in this point but the definition should still match the standard.
- [x] Edit the expected test results accordingly. Due to less precise gas constant for dry air and 50% humidity.

I'll open a pull request on this issue soon and also change the documentation accordingly.

PR #11, opened 2019-04-30
Adjust air density calculation to the IEC 61400-12 definition

sebastianpfaffel:
Solves the issues stated in #10. The calculation of air density is now equal to the IEC 61400-12 definition.  The unit test fails till the expected results are modified as well. 

- Air density calculations considers now humidity and use a default value of 0.5 in case of missing values. 
- The gas constant for dry air is changed to the IEC standard of 287.05 instead of 287.058. The IEC definition is less precise in this point but the definition should still be streamlined with the standard.
- Modified the expected test results accordingly. Due to less the precise gas constant for dry air and 50% humidity.

### DIFF
--- a/operational_analysis/toolkits/met_data_processing.py
+++ b/operational_analysis/toolkits/met_data_processing.py
@@ -45,7 +45,7 @@ def compute_u_v_components(wind_speed, wind_dir):
     return u, v
 
 
-def compute_air_density(df, temp_col, pres_col, humi_col = 'default'):
+def compute_air_density(df, temp_col, pres_col, humi_col = None):
     """
     Calculate air density from the ideal gas law based on the definition provided by IEC 61400-12
     given pressure, temperature and relative humidity.
@@ -63,14 +63,13 @@ def compute_air_density(df, temp_col, pres_col, humi_col = 'default'):
         :obj:`pandas.Series`: Rho, calcualted air density; units of kg/m3
     """
     # Check if humidity column is provided and create default df (0.5) if necessary
-    if humi_col in df:
+    if humi_col in df and humi_col != None:
         rel_humidity = df.loc[:, [humi_col]]
     else:
+        humi_col = 'relative_humidity'
         rel_humidity = pd.DataFrame([0.5]*df[temp_col].shape[0],columns=[humi_col], index=df.index)
-        df['default'] = [0.5]*df[temp_col].shape[0]
     # Send exception if any negative data found
-    if (df.loc[df[temp_col] < 0].shape[0] > 0) | (df.loc[df[pres_col] < 0].shape[0] > 0 
-       | (rel_humidity.loc[rel_humidity[humi_col] < 0].shape[0] > 0)):
+    if (df[temp_col] < 0).any() | (df[pres_col] < 0).any() | (rel_humidity[humi_col] < 0).any():
         raise Exception('Some of your temperature, pressure or humidity data is negative. Check your data.')
 
     R_const = 287.05  # Gas constant for dry air, units of J/kg/K
--- a/operational_analysis/toolkits/met_data_processing.py
+++ b/operational_analysis/toolkits/met_data_processing.py
@@ -50,7 +50,7 @@ def compute_air_density(df, temp_col, pres_col, humi_col = 'default'):
     Calculate air density from the ideal gas law based on the definition provided by IEC 61400-12
     given pressure, temperature and relative humidity.
     This function assumes temperature and pressure are reported in standard units of measurement
-    (i.e. Kelvin for temperature, Pascal for pressure, humidity has no dimension)
+    (i.e. Kelvin for temperature, Pascal for pressure, humidity has no dimension).
     Humidity values are optional. According to the IEC a humiditiy of 50% (0.5) is set as default value. 
 
     Args:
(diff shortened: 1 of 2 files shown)


operational_analysis/toolkits/met_data_processing.py
```diff
@@ -45,26 +45,38 @@ def compute_u_v_components(wind_speed, wind_dir):
...
 
     Returns:
         :obj:`pandas.Series`: Rho, calcualted air density; units of kg/m3
     """
+    # Check if humidity column is provided and create default df (0.5) if necessary
+    if humi_col in df:
+        rel_humidity = df.loc[:, [humi_col]]
+    else:
+        rel_humidity = pd.DataFrame([0.5]*df[temp_col].shape[0],columns=[humi_col], index=df.index)
+        df['default'] = [0.5]*df[temp_col].shape[0]
```
jordanperr: Setting this column of the input data frame may cause unintended side effects. Let's say df contains an unrelated column "default", and humi_col is set to "foo", which is not present in df. The else branch will then be taken here, and "default" will be overwritten. I suggest keeping this "default" column as a series (with a more descriptive name!) instead of adding it to the input dataframe.

operational_analysis/toolkits/met_data_processing.py
```diff
@@ -45,26 +45,38 @@ def compute_u_v_components(wind_speed, wind_dir):
...
     # Send exception if any negative data found
-    if (df.loc[df[temp_col] < 0].shape[0] > 0) | (df.loc[df[pres_col] < 0].shape[0] > 0):
-        raise Exception('Some of your temperature of pressure data is negative. Check your data')
-
-    R_const = 287.058  # Gas constant for dry air, units of J/kg/K
-    rho = df[pres_col] / df[temp_col] / R_const
+    if (df.loc[df[temp_col] < 0].shape[0] > 0) | (df.loc[df[pres_col] < 0].shape[0] > 0 
+       | (rel_humidity.loc[rel_humidity[humi_col] < 0].shape[0] > 0)):
```
jordanperr: I understand you were simply extending this pattern, but may be more succinct to use df[temp_col]<0).any() etc...

operational_analysis/toolkits/met_data_processing.py
```diff
@@ -45,26 +45,38 @@ def compute_u_v_components(wind_speed, wind_dir):
...
+        raise Exception('Some of your temperature, pressure or humidity data is negative. Check your data.')
+
+    R_const = 287.05  # Gas constant for dry air, units of J/kg/K
+    Rw_const = 461.5   # Gas constant of water vapour, unit J/kg/K
+    rho = ((1/df[temp_col])*(df[pres_col]/R_const-rel_humidity[humi_col]*(0.0000205*np.exp(0.0631846*df[temp_col]))*
```
jordanperr: What are the two constants embedded inline on 78? 0.0000205 and 0.0631846.

Can we pull these constants from somewhere else? Is thee a standard package with these? I could support the creation of a constants module.

moptis:
I think the science looks good and conforms to the IEC standard.

@sebastianpfaffel We have talked internally about moving away from the data frame approach in this function and making it more generic. You'll notice some functions in this met data processing toolkit accept individual series as inputs (e.g. pressure_vertical_extrapolation), and some require data frames. I think the latter is too restrictive and we should be pushing towards a series input approach for all functions in this toolkit. 

@jordanperr I think this adjustment would address your concern about unintended side effects. 

@sebastianpfaffel Do you want to adjust this function to the series input format? If not, I can take what you did and submit an update to this pull request.

jordanperr:
> We have talked internally about moving away from the data frame approach in this function and making it more generic.

While I agree in principle, in practice this would be a breaking API change to this function which may have consequences in other parts of the codebase - or in other codebases that rely on OpenOA. It also seems out of scope for issue #10. I recommend merging this pull request to resolve the issue and then opening a new ticket to switch the met_toolkit to the Pandas Series api.

operational_analysis/toolkits/met_data_processing.py
sebastianpfaffel: Thanks for the review. To be honest, line 70 shouldn't be existing in the pull request at all and it's not of use any in the script. I created the new data frame "rel_humidity" to avoid this issue. Line 70 can be simply deleted. The naming issue is still existing. The default value for humi_col should be probably an empty ''  I'll change that when I'm back home on Monday.

operational_analysis/toolkits/met_data_processing.py
sebastianpfaffel: Sounds reasonable, I'll edit this when applying the other changes.

sebastianpfaffel:
I'm happy to go either way as long as we keep it consistent throughout the framework. You probably want to decide this internally. Otherwise we can discuss it during the conference call next week as well. IMHO the series approach seems to be more flexible and would be make it easier to reuse the code. 

@moptis I can adjust the function as soon as we have a decision on this issue.  Feel free to update the pr anytime though.

operational_analysis/toolkits/met_data_processing.py
sebastianpfaffel: Even though the constants do probably exist in a standard package, the precision of the IEC definition might differ to other sources. The constants in line 78 aren't physical constants but simply a best fit in the equation for vapor pressure and thus hardcoded.

sebastianpfaffel:
I've updated the PR according to your comments and our discussion. Feel free to review, correct and merge.

jordanperr:
Looks good to me. Thanks, Sebastian!

Issue #17, opened 2019-05-16
Use ENGIEs open data windfarm as example data set


sebastianpfaffel:
The ENGIE open data windfarm seems to be a good data set to test and demonstrate the capabilities of OpenOA while using publicly available data. Thus, I'd like to add an example of how to import turbine data from this source into OpenOA and create an additional class similar to the TurbineExampleProject class. 

Furthermore, I'd either like to edit the "Turbine Level Toolkit Notebook" example in the docs to use the ENGIE data instead of the CSV file or just add the info of hot to use it as an alternative data source to the same notebook. Any opinions on that?

moptis:
@sebastianpfaffel  I am in support of this idea. Would the Engie data need to be hosted here, or can it be accessed through an API? The latter would be obviously advantageous.

sebastianpfaffel:
@moptis Thanks for your reply. The data is hosted by Engie and available on https://opendata-renewables.engie.com/ for a manual download of through an HTTP REST API. Thus it can be directly loaded by a python script and doesn't need to be hosted by OpenOA. I have some first code ready and will create a PR as soon as I have examples ready as well.

sebastianpfaffel:
I finally found some time to finish the import script and edit the example in the ipnb. Would be great if someone can review it. We could get rid of the csv file, but someone would need to find a period with more outliers in the ENGIE data. #22

sebastianpfaffel:
Thanks for your review and merging!

PR #22, opened 2019-06-28
#17 Add the ENGIE open data wind farm to the examples

sebastianpfaffel:
#17 

I added the possibility to create turbine objects based on the ENGIE open data wind farm in the examples. An example was added to the ipnb as well. I had to fix some other minor issues on the ipnb as well to make it work. 

- print ("xyz") instead of print "xyz", Python 2.x vs. Python 3.x
- gam(...) instead of spline(....), the spline function has been renamed or deleted?


jordanperr:
@moptis and @nateagarwal are updating turbine_gross_energy and the turbine example project to use -25 column names. I am waiting to merge in their pull request #28 before updating this one to reflect the new column names. Aside from this small change then PR looks good to me.

sebastianpfaffel:
@jordanperr Thanks for your feedback. Let me know if there is anything else I can do or if any further changes are needed.

jordanperr:
I've initiated a pull request into your develop branch. Please review my changes and merge them at your convenience. Once that is merged - I will be comfortable merging this PR into develop.

sebastianpfaffel:
Thanks for the PR. It makes sense to split the examples and I merged it. Next time I'll make use of a separate branch for that issue again, kinda missed ... ;)  Do we need to change the docs to consider both examples now? Changes could be done in #29.

jordanperr:
Yeah - I think it'd be better to implement #29 so the notebook is automatically included in the documentation. Up to you, though. I'm fine either way. Not a big deal to have an extra example notebook that is not mentioned in the documentation for now.

Issue #26, opened 2019-07-05
Turbine gross energy method - remove NaNs from merged data frame


moptis:
In setup_model_dict, the daily energy data are merged with the daily average reanalysis wind speeds. Some NaNs may be produced if reanalysis data do not cover some timestamps in the energy data. Add a simple dropna() line to remove all NaNs that may occur during this merge.

jordanperr:
Has this issue been closed with PR #28?

PR #28, opened 2019-07-05
Nan check on turb gross, fix example column names

moptis:
Address Issue #26 

Small changes to turb gross energy method to drop NaNs from merged energy and reanalysis daily data frame. Also updated the example import to be consistent with -25.


moptis:
I updated the unit_conversion test to reflect our previous PR (i.e., taking datetime frequency rather than an integer frequency), and added more frequency options to the power_to_enegy method to reflect those used in the unit_conversion test function.

moptis:
@jordanperr  Very short PR here. Got a sec to review today? Bit of a push to get gap analysis fully working next week. Let me know what you can do!

nateagarwal:
@moptis @jordanperr I just tested this pull request with two projects and it addresses the labeled issue.

jordanperr:
Changing the column names in turbine_project.py invalidates the example notebook, and therefore the RST file based on that  notebook. These files will have to be updated to reflect the new column names and re-run:

https://github.com/NREL/OpenOA/blob/master/examples/turbine_analysis/Turbine_Toolkit_Examples.ipynb
https://github.com/NREL/OpenOA/blob/master/sphinx/Turbine_Toolkit_Examples.rst

jordanperr:
I have updated the notebook and rst files within this PR. Unfortunately, I also had to delete the spline power curve fit from the turbine example, since we no longer have a spline fit in the power curves module.

PR #30, opened 2019-07-09
Feature/elec losses

moptis:
Adding electrical loss method for a wind plant. 

Basic approach is to calculate daily sum of turbine energy and meter energy, isolating only the days where all timestamps are reporting from both data sources. Electrical loss is then the difference between the sum of turbine vs. sum of meter energy over those concurrent days.

In the case where only monthly metered data is provided (i.e. daily sums not possible), we instead correct the daily sum of turbine energy for any unreported data. These corrected energy values are then summed on a monthly basis and compared to the metered monthly energy to assess electrical loss.


operational_analysis/methods/turbine_long_term_gross_energy.py
```diff
@@ -60,7 +60,6 @@ def __init__(self, plant, max_power_filter = 0.85, wind_bin_thresh=2):
         
         self._max_power_filter = max_power_filter # Parameter used for bin-based filtering
         self._wind_bin_thresh = wind_bin_thresh
-        self._reanal = ['merra2', 'erai', 'ncep2'] # Reanalysis products to consider
```
moptis: Sorry, looks like an improvement to turbine gross energy snuck in here. The idea is to not necessarily run all 3 reanalysis products but, similar to the Monte Carlo AEP method, allow the user to specify which reanalysis products to run.

operational_analysis/methods/turbine_long_term_gross_energy.py
```diff
@@ -83,7 +82,7 @@ def __init__(self, plant, max_power_filter = 0.85, wind_bin_thresh=2):
         
 
     @logged_method_call
-    def run(self):
+    def run(self, reanal_subset):
```
moptis: reanalysis subset is now a parameter, similar to Monte Carlo AEP.

moptis:
@jordanperr  It doesn't appear that Travis CI ran here?

jordanperr:
It looks like it did run, and passed. https://travis-ci.org/NREL/OpenOA/builds/557928257
I am not sure why it's not updating here in Github's UI. I am going to try and open a support ticket with Travis-CI.

PR #32, opened 2019-07-16
Feature/turb gross fixes

moptis:
Several updates to the turbine long term gross energy method to make it more accurate and robust. Here's a summary of the updates:

1. In order to preserve more data when building stats models between daily turbine energy sum and daily average reanalysis wind speed, I've done two things:  

a) I've relaxed the requirement that all timestamps must be reporting (e.g. 144 counts per day for 10-min data) and instead allow for 90%+ reporting. If less than 100% reporting, energy is corrected based on the ratio of expected to actual data count (i.e. a time-based energy correction).

b) For remaining days that don't make the above cut, daily energy is imputed from other correlated turbines

2. I've added plotting options to the run() method which produces flagged turbine data plots at native resolution (e.g., 10-min) and also plots of the daily energy and reanalysis averaged wind speed alongside the long-term estimated energy.


operational_analysis/methods/turbine_long_term_gross_energy.py
```diff
@@ -45,7 +47,7 @@ class TurbineLongTermGrossEnergy(object):
     """
 
     @logged_method_call
-    def __init__(self, plant, max_power_filter = 0.85, wind_bin_thresh=2):
+    def __init__(self, plant):
```
moptis: Moved these parameters to the run() method.

operational_analysis/methods/turbine_long_term_gross_energy.py
```diff
@@ -58,17 +60,19 @@ def __init__(self, plant, max_power_filter = 0.85, wind_bin_thresh=2):
         self._plant = plant  # Set plant as attribute of analysis object
         self._turbs = self._plant._scada.df['id'].unique() # Store turbine names
         
-        self._max_power_filter = max_power_filter # Parameter used for bin-based filtering
-        self._wind_bin_thresh = wind_bin_thresh
+        # Get start and end of POR days in SCADA
+        self._por_start = format(plant._scada.df.index.min(), '%Y-%m-%d')
+        self._por_end = format(plant._scada.df.index.max(), '%Y-%m-%d')
+        self._full_por = pd.date_range(self._por_start, self._por_end, freq = 'D')
```
moptis: Define full period of record for SCADA to be used in imputing later.

operational_analysis/methods/turbine_long_term_gross_energy.py
```diff
@@ -77,30 +81,53 @@ def __init__(self, plant, max_power_filter = 0.85, wind_bin_thresh=2):
...
+
+        
         # Initially sort the different turbine data into dictionary entries
+        logger.info("Processing SCADA data into dictionaries by turbine (this can take a while)")
         self.sort_scada_by_turbine()
         
 
     @logged_method_call
-    def run(self, reanal_subset):
+    def run(self, reanal_subset = ['erai', 'ncep2', 'merra2'], max_power_filter = 0.85,
```
moptis: Add more parameters to the run() method, including plotting indicators and save file

operational_analysis/methods/turbine_long_term_gross_energy.py
```diff
@@ -77,30 +81,53 @@ def __init__(self, plant, max_power_filter = 0.85, wind_bin_thresh=2):
...
         self.filter_turbine_data() # Filter turbine data
         
+        if enable_plotting:
+            logger.info("Plotting filtered power curves")
+            self.plot_filtered_power_curves(plot_dir) # Setup daily reanalysis products
+        
         logger.info("Processing reanalysis data to daily averages")
         self.setup_daily_reanalysis_data() # Setup daily reanalysis products
         
+        logger.info("Processing scada data to daily sums")
```
moptis: I split out processing of daily scada from setup_model_dict() into its own function since more steps are now included (e.g. imputing)

jordanperr:
Todo:
- Add test case with missing data
- Make sure docs are up to date (reflecting new method)

HOLD

PR #41, opened 2019-08-22
Feature/lt por ratio calc

moptis:
Made changes to plant_analysis to support calculation of the long-term / period of record ratio. I also modified figures for public dissemination and sped up the MC code by replacing a dataframe with numpy arrays.

### DIFF
--- a/operational_analysis/methods/plant_analysis.py
+++ b/operational_analysis/methods/plant_analysis.py
@@ -220,6 +220,29 @@ class MonteCarloAEP(object):
 
         return plt
 
+    def plot_aep_boxplot(self, param, lab):
+        """                                                                                                                                                                                        
+        Plot box plots of AEP results sliced by a specified Monte Carlo parameter                                                                                                                  
+
+        Args:                                                                                                                                                                                      
+           param( :obj:`list'): The Monte Carlo parameter on which to split the AEP results                                                                                                                   lab(:obj:'str'): The name to use for the parameter when producing the figure
+        
+        Returns:                                                                                                                                                                                   
+            (none)                                                                                                                                                                               
+        """
+
+        import matplotlib.pyplot as plt
+        sim_results = self.results
+
+        tmp_df=pd.DataFrame(data={'aep': sim_results.aep_GWh, 'param': param})
+        tmp_df.boxplot(column='aep',by='param',figsize=(8,6))
+        plt.ylabel('AEP (GWh/yr)')
+        plt.xlabel(lab)
+        plt.title('AEP estimates by %s' % lab)
+        plt.suptitle("")
+
+        return plt
+
     def plot_monthly_plant_data_timeseries(self):
         """
         Plot timeseries of monthly gross energy, availability and curtailment
--- a/operational_analysis/methods/plant_analysis.py
+++ b/operational_analysis/methods/plant_analysis.py
@@ -351,7 +351,6 @@ class MonteCarloAEP(object):
         if (self._plant._meter_freq == '1MS') | (self._plant._meter_freq == '1M'):
             self._monthly.df['num_days_actual'] = self._monthly.df['num_days_expected']            
         else:
-            print('yes')
             self._monthly.df['num_days_actual'] = df.resample('MS')['energy_kwh'].apply(tm.num_days)
 
     @logged_method_call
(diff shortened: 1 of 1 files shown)


operational_analysis/methods/plant_analysis.py
```diff
@@ -75,9 +75,19 @@ def __init__(self, plant, uncertainty_meter=0.005, uncertainty_losses=0.05,
         
         self.num_days_lt= (31, 28.25, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31)
 
-        # Run preprocessing step
+        # Run preprocessing step                                                                                                                                                                                   
         self.calculate_monthly_dataframe()
 
+        # Store start and end of period of record
```
moptis: Record start and end dates for plant period of record. Record these dates in the reanalysis dataframe

operational_analysis/methods/plant_analysis.py
```diff
@@ -335,10 +348,11 @@ def process_revenue_meter_energy(self):
         # Get actual number of days per month in the raw data
         # (used when trimming beginning and end of monthly data frame)
         # If meter data has higher resolution than monthly
-        if (self._plant._meter_freq != '1MS') & (self._plant._meter_freq != '1M'):
-            self._monthly.df['num_days_actual'] = df.resample('MS')['energy_kwh'].apply(tm.num_days)
+        if (self._plant._meter_freq == '1MS') | (self._plant._meter_freq == '1M'):
```
moptis: Didn't like how this was set up (similar to dissatisfaction with range_flag), so I adjusted. Same effect.

operational_analysis/methods/plant_analysis.py
```diff
@@ -655,28 +669,39 @@ def run_AEP_monte_carlo(self):
 
         num_sim = self.num_sim
 
-        sim_results = pd.DataFrame(index=np.arange(num_sim), data={'aep_GWh': np.empty(num_sim),
-                                                                   'avail_pct': np.empty(num_sim),
-                                                                   'curt_pct': np.empty(num_sim)})
+        aep_GWh = np.empty(num_sim)
```
moptis: Improve simulation speed by ~25% by using numpy arrays instead of pandas dataframe.

operational_analysis/methods/plant_analysis.py
```diff
@@ -655,28 +669,39 @@ def run_AEP_monte_carlo(self):
...
         for n in tqdm(np.arange(num_sim)):
             slope, intercept = self.run_regression(n)  # Get slope, intercept from regression
             ws_lt = self.sample_long_term_reanalysis(self._mc_num_years_windiness[n],
                                                      self._mc_reanalysis_product[n])  # Get long-term wind speeds
+            
             # Get long-term normalized gross energy by applying regression result to long-term monthly wind speeds
             gross_norm_lt = ws_lt.multiply(slope) + intercept
             gross_lt=gross_norm_lt*self.num_days_lt/30 # Undo normalization to 30-day months
             
+            # Get long-term normalized gross energy by applying regression result to long-term monthly wind speeds
```
moptis: Part of long-term / POR ratio calculation

operational_analysis/methods/plant_analysis.py
```diff
@@ -655,28 +669,39 @@ def run_AEP_monte_carlo(self):
...
-            sim_results.loc[n,'aep_GWh'] = gross_lt.sum() * (1 - avail_lt_losses)
-            sim_results.loc[n,'avail_pct'] = avail_lt_losses
-            sim_results.loc[n,'curt_pct'] = curt_lt_losses
+            aep_GWh[n] = gross_lt.sum() * (1 - avail_lt_losses)
+            avail_pct[n] = avail_lt_losses
+            curt_pct[n] = curt_lt_losses
+            lt_por_ratio[n] = gross_lt.sum() / gross_por.sum()            
 
         # Return final output
+        sim_results = pd.DataFrame(index=np.arange(num_sim), data={'aep_GWh': aep_GWh,
```
moptis: Now combine in data frame after looping through MC simulations

jordanperr:
Looks good to me. Interesting that col[n] = array is so much faster than df.loc[n,"col"] = array. Dataframe.loc uses labels, while iloc uses indexes. This makes me think you could achieve a similar speed by first dereferencing the column using a label lookup. But I think using numpy arrays makes sense for the code as its written now.

PR #44, opened 2019-09-04
Add duplicate index filter

sebastianpfaffel:
I added a filter to remove duplicate index values from the data. This is a common problem if your data includes a switch between DST and summertime. The engie example was updated accordingly. 

Furthermore, I changed  "np.repeat("True", engie.scada.df.shape[0])" to "np.repeat(False, engie.scada.df.shape[0])" in both examples. There should be no reason why strings are used when the functions expects boolean values.


jordanperr:
Hi Sebastian,

Thanks for this pull request. Upon review, I'm not certain another duplicate timestamp function is required. We already have a function in the timeseries module called find_duplicate_times: https://github.com/NREL/OpenOA/blob/develop/operational_analysis/toolkits/timeseries.py#L71

Additionally, I'm not convinced we really need any of these wrapping functions. Why not just use pd.duplicated directly in the notebook?

Jordan

jordanperr:
Hi Sebastian, I will close this pull request shortly since there hasn't been any activity. Please let me know if you'd like me to keep it open. Thanks!

sebastianpfaffel:
Hi Jordan, 
sorry for my inactivity lately. I somehow missed the already existing duplicates filter. 

I closed the PR and will open a separate PR to update the example on the engie_data at some point. 

Sebastian

PR #47, opened 2019-09-13
Automated QC Method and Updated Engine Project for Tutorial

nateagarwal:
This PR represents a first pass at the QC process for an Engie-based tutorial series. 

The main file changes include:
-quality_check_automation.py: This file contains the methods necessary for a QC process with documentation. Note that certain methods in this file are currently relying upon the Wind Toolkit. This dependance will need to be handled in a later PR.
-Engie_turbine_QC.ipnb: In this notebook, a sample QC process is presented for a subset of Engie turbine data. Note that the lat/lon provided in this notebook is inconsistent with reality in order to avoid challenges with the WTK dependance. 
- project_engie_scada.py: This represents a first pass at an import script for the Engie project to prepare for turbine-level and plant-level analysis.


moptis:
@nateagarwal This is looking great. Have you tried running this for various test cases (e.g., meter df, scada df, etc.). I downloaded this branch and tried running t75 meter data as an example and I'm getting a fail. Looks like you're assuming the index is datetime but may not necessarily be so. Let's make sure this can run on a bunch of different Phase 1 datasets before merging.
  
image

nateagarwal:
@moptis this code has two implicit assumptions:
- The date time field is set as the index
-The id field is *only* provided for scada data. 

Happy to explore options to make this more accessible to different datasets.

moptis:
@nateagarwal I like the idea of not assuming index is datetime. Otherwise, why would we need to specify the datetime column in the init() function? Seems duplicative. In general, we've approached QC in those notebooks not making any assumptions about what the index is. I'd like to see that continue here if possible. It shouldn't change too much code I wouldn't think.

nateagarwal:
@moptis I have made the appropriate index generalization changes

jordanperr:
Edit comment by @moptis:

@nateagarwal This is looking great. Have you tried running this for various test cases (e.g., meter df, scada df, etc.). I downloaded this branch and tried running t75 meter data as an example and I'm getting a fail. Looks like you're assuming the index is datetime but may not necessarily be so. Let's make sure this can run on a bunch of different Phase 1 datasets before merging.

operational_analysis/methods/quality_check_automation.py
```diff
@@ -0,0 +1,345 @@
...
+            df_diurnal = self._df.groupby(self._df[self._t].dt.hour)[self._w].mean()
+
+        return_corr = np.empty((24))
+        for i in np.arange(24):
+            df_temp = df_diurnal.shift(i)
+        
+            if i != 0:
+                df_temp[np.arange(i)] = df_diurnal[-i:]
+
+            return_corr[i] = np.corrcoef(self._ws_diurnal['ws'], df_temp)[0,1]
```
jordanperr: How about np.roll? I reckon lines 212-219 can be replaced with:

return_corr = np.empty((24))
for i in np.arange(24):
    return_corr[i] = np.corrcoef(self._ws_diurnal['ws'], np.roll(df_diurnal["ws"],i))[0,1]

operational_analysis/methods/quality_check_automation.py
```diff
@@ -0,0 +1,345 @@
...
+        dst_df = pd.read_csv('./daylight_savings.csv')
+
+        if self._id != 'None':
+            self._df_dst =  self._df.groupby(self._df[self._t])[self._w].sum().to_frame()
+        else:
+            self._df_dst = self._df
+    
+        self._df_dst['time'] = self._df_dst[self._t]
+        df_full = timeseries.gap_fill_data_frame(self._df_dst, 'time', self._freq) # Gap fill so spring ahead is visible
+        df_full.set_index('time', inplace = True) # Have to reset index to datetime
```
jordanperr: Is this needed because timeseries.gap_fill_data_frame is resetting the index? Or are we setting the index for the first time here? If it's the later, what is the meaning of the inline comment?

operational_analysis/methods/quality_check_automation.py
```diff
@@ -0,0 +1,345 @@
...
+        plt.grid()
+        plt.xlabel('Hour of day')
+        plt.ylabel('Normalized values')
+        plt.title('WTK and QC Timezone Comparison')
+        plt.legend()
+        plt.show()
+
+    def corr_df_calc(self):
+        """
+        This method returns a correlation series that compares the current power data (with different shift thresholds) to wind speed data from the WTK with hourly resolution.
```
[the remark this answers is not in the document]
nateagarwal: Addressed

operational_analysis/methods/quality_check_automation.py
```diff
@@ -0,0 +1,345 @@
...
+from operational_analysis.toolkits import timeseries
+from operational_analysis.toolkits import filters
+from operational_analysis.toolkits import power_curve
+
+from operational_analysis import logged_method_call
+from operational_analysis import logging
+
+logger = logging.getLogger(__name__)
+
+class QCAuto(object):
```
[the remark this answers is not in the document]
nateagarwal: Addressed

operational_analysis/methods/quality_check_automation.py
```diff
@@ -0,0 +1,345 @@
...
+        Args:
+        (None)
+
+        Returns:
+        (None)
+        """
+        self._time_gaps = timeseries.find_time_gaps(self._df[self._t], freq=self._freq)
+
+    def indicesForCoord(self,f):
+
```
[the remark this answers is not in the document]
nateagarwal: Addressed

operational_analysis/methods/quality_check_automation.py
```diff
@@ -0,0 +1,345 @@
...
+        Returns:
+            (None)
+        """
+
+        logger.info("Identifying Time Duplications")
+        self.dup_time_identification()
+        logger.info("Identifying Time Gaps")
+        self.gap_time_identification()
+        logger.info("Evaluating timezone deviation from UTC")
+        self._ws_diurnal = self.ws_diurnal_prep()
```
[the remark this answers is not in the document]
nateagarwal: Addressed

operational_analysis/methods/quality_check_automation.py
```diff
@@ -0,0 +1,345 @@
...
+            scada_sub = self._df.loc[self._df[self._id] == t, :]
+            plt.scatter(scada_sub[self._ws], scada_sub[self._w], s = 5)
+            n = n + 1
+            plt.title(t)
+        plt.tight_layout()
+        plt.show()
+
+    def column_plots(self):
+        """
+        Produces plots for each numeric column.
```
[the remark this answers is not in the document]
nateagarwal: Addressed

operational_analysis/methods/quality_check_automation.py
```diff
@@ -0,0 +1,345 @@
...
+        """
+
+        self._max_min = pd.DataFrame(index = self._df.columns, columns = {'max', 'min'})
+        self._max_min['max'] = self._df.max()
+        self._max_min['min'] = self._df.min()
+
+    def turb_plots(self):
+
+        """
+        Produces plots of each individual turbine
```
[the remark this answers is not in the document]
nateagarwal: Addressed

operational_analysis/methods/quality_check_automation.py
nateagarwal: @jordanperr I don't think that time series.gap_fill_data_frame sets the index (but it does sort by time). I have removed the inline comment; it is an artifact from removing the time index dependence for the main DataFrame.

operational_analysis/methods/quality_check_automation.py
nateagarwal: Applied

operational_analysis/methods/quality_check_automation.py
```diff
@@ -0,0 +1,345 @@
...
+
+        Args:
+        hour_window(:obj: 'int'): number of hours outside of the Daylight Savings Time transitions to view in the plot (optional)
+
+        Returns:
+        None
+        """
+    
+        # List of daylight savings days back to 2010
+        dst_df = pd.read_csv('./daylight_savings.csv')
```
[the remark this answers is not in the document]
nateagarwal: @jordanperr this is a good suggestion and I am exploring this. Perhaps there could be an added parameter, local_time_zone, and with this, two plots could be created: one for UTC dates and another for local time?

operational_analysis/methods/quality_check_automation.py
```diff
@@ -0,0 +1,345 @@
...
+        This method returns a correlation series that compares the current power data (with different shift thresholds) to wind speed data from the WTK with hourly resolution.
+
+        Args:
+        (None)
+
+        Returns:
+        (None)
+        """
+        if self._id != 'None':
+            scada_sum = self._df.groupby(self._df[self._t])[self._w].sum().to_frame()
```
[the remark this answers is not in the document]
nateagarwal: @jordanperr this groupby feature is specifically introduced for the turbine data. We want to look at a sum of all the power contributions from individual turbines for a given timestamp.

operational_analysis/methods/quality_check_automation.py
nateagarwal: @jordanperr Re-evaluating this, I see your point. In addressing other changes, the double index-setting becomes apparent.

nateagarwal:
@moptis I have tested this qc method with four projects (two external and two internal cases). I would like to flag one internal case in which the daylight savings plots were inconsistent with previous analysis and the hour correlations demonstrated a similar trend but not the same exact values. 

@jordanperr I have addressed your comments, with the exception of pytz integration. I am happy to integrate pytz as other comments are provided.

operational_analysis/methods/quality_check_automation.py
```diff
@@ -0,0 +1,345 @@
...
+        """
+        This function returns any time duplications in the dataset.
+
+        Args:
+        (None)
+
+        Returns:
+        (None)
+        """
+        if self._id!= 'None':
```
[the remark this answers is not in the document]
nateagarwal: Addressed

nateagarwal:
@moptis this most recent commit reconciles the internal consistency of results.

operational_analysis/methods/quality_check_automation.py
```diff
@@ -0,0 +1,331 @@
...
+        coords = (lon, lat)
+        coords = projectLcc(*coords)
+        delta = np.subtract(coords, origin)
+        ij = [int(round(x/2000)) for x in delta]
+        return tuple(reversed(ij))
+
+    def ws_diurnal_prep(self, start_date = '2007-01-01', end_date = '2013-12-31'):
+
+        """
+        This method returns a Pandas Series corresponding to hourly average windspeeds
```
moptis: This does a lot more than just return a pandas series. They key is that It links into the WTK data on AWS as the data source, grabs wind speed data, and calculates diurnal hourly averages.  I think it'd be useful to expand this docstring to include a more detailed description.

operational_analysis/methods/quality_check_automation.py
```diff
@@ -0,0 +1,331 @@
...
+    
+        # Calculate diurnal profile of wind speed
+        ws_diurnal=ws_df.groupby(ws_df.index.hour).mean()
+
+        return ws_diurnal
+
+    def wtk_diurnal_plot(self):
+
+        """
+        This method plots the WTK diurnal plot alongisde the hourly power averages of the scada_df across all turbines
```
moptis: Not necessarily scada_df, could be any power or wind speed data, no?

operational_analysis/methods/quality_check_automation.py
```diff
@@ -0,0 +1,331 @@
...
+        Args:
+        (None)
+        Returns: (None)
+        """
+
+        sum_df = self._df.groupby(self._df[self._t])[self._w].sum().to_frame()
+        df_temp = sum_df.copy()
```
moptis: Why do we need to copy?

operational_analysis/methods/quality_check_automation.py
```diff
@@ -0,0 +1,331 @@
...
+
+        Returns:
+        (None)
+        """
+
+        self._max_min = pd.DataFrame(index = self._df.columns, columns = {'max', 'min'})
+        self._max_min['max'] = self._df.max()
+        self._max_min['min'] = self._df.min()
+
+    def turb_plots(self):
```
moptis: Can we call this something more intiuitive (e.g. power_curve_plot). After all it could be revenue meter data we're plotting.

operational_analysis/methods/quality_check_automation.py
nateagarwal: Addressed

operational_analysis/methods/quality_check_automation.py
```diff
@@ -0,0 +1,331 @@
...
+
+        """
+        Produces power curve (power vs. wind speed) plot for each individual turbine
```
moptis: Revenue meter data as well?

operational_analysis/methods/quality_check_automation.py
nateagarwal: I am removing the setting/copy error so that it's clear that we don't just want a view of the DataFrame.

operational_analysis/methods/quality_check_automation.py
```diff
@@ -0,0 +1,331 @@
...
+        for t in turbs:
+            plt.subplot(num_rows, 4, n)
+            scada_sub = self._df.loc[self._df[self._id] == t, :]
+            plt.scatter(scada_sub[self._ws], scada_sub[self._w], s = 5)
+            n = n + 1
+            plt.title(t)
+        plt.tight_layout()
+        plt.show()
+
+    def column_plots(self):
```
moptis: Not sure what column plot means. We're plotting histograms for every data field in a df, right?

operational_analysis/methods/quality_check_automation.py
nateagarwal: @moptis this function specifically handles turbine plots and will fail if you try to run it for meter data. The column_plots function plots histograms for any numeric column.

I would be more than happy to change the method name; I agree that it is more intuitive.

operational_analysis/methods/quality_check_automation.py
nateagarwal: Yes; every numeric column (i.e. int or float). I can change the name to be more intuitive.

operational_analysis/methods/quality_check_automation.py
```diff
@@ -0,0 +1,331 @@
...
+        plt.plot(ws_norm, label = 'WTK wind speed (UTC)')
+        plt.plot(df_norm, label = 'QC power')
+        plt.grid()
+        plt.xlabel('Hour of day')
+        plt.ylabel('Normalized values')
+        plt.title('WTK and QC Timezone Comparison')
+        plt.legend()
+        plt.show()
+
+    def corr_df_calc(self):
```
moptis: This seems to print out the correlation series in the reverse order (i.e. highest correlations in the 19, 20 range, and not the 4 or 5 range. I think it's more intuitive to have the latter (i.e., looks like a 5 hour shift to get the data into UTC).

nateagarwal:
@moptis this commit addresses your comments.

moptis:
@jordanperr I'm good with this as is. What do you think? Still some improvements to be made and no unit or integration tests yet. Thoughts?

operational_analysis/methods/quality_check_automation.py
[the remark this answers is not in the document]
jordanperr: Ahh, so it assumes the turbine data is time aligned (each timestamp will be present across all turbines) and therefore would be multiple entries with the same timestamp in the dataframe. That makes sense.

operational_analysis/methods/quality_check_automation.py
jordanperr: I don't think this was actually applied in the code. But it's not a blocker.

operational_analysis/methods/quality_check_automation.py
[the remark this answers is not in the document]
nateagarwal: @jordanperr Yep. However, by making your change of introducing an id column for all data, I generalized this (and many other) functions. Good tip!

operational_analysis/methods/quality_check_automation.py
```diff
@@ -0,0 +1,345 @@
...
+        (None)
+
+        Returns:
+        (None)
+        """
+        if self._id != 'None':
+            scada_sum = self._df.groupby(self._df[self._t])[self._w].sum().to_frame()
+            df_diurnal = scada_sum.groupby(scada_sum[self._t].dt.hour)[self._w].mean()
+        else:
+            df_diurnal = self._df.groupby(self._df[self._t].dt.hour)[self._w].mean()
```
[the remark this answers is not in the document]
nateagarwal: Yep

operational_analysis/methods/quality_check_automation.py
nateagarwal: @jordanperr This change was originally made, and @moptis requested that the original order be preserved.

PR #57, opened 2019-12-09
Joss paper

jordanperr:
Pull request for the paper we plan to submit to JOSS: http://joss.theoj.org

Outstanding issues:
- Should add citations to WP3 work
- Over word limit by 200 words (limit is 1,000 words)

You can compile this paper using an online preview service offered by the journal:
https://whedon.theoj.org

During the development of the pull request, you can use the following inputs:
Paper repository address: https://github.com/jordanperr/OpenOA
Custom branch (optional): joss-paper


jordanperr:
@moptis I don't see your changes in my branch. Can you verify they were pushed? The commits will show up here once you push to jordanperr/OpenOA

jordanperr:
Comments for @moptis 's changes:
- Travis and anna were deleted from the author list.
- Single line breaks do not matter in markdown.

@Dynorat and I discussed whether or not the structure of this paper should be changed to reflect the new toolkits, but decided against doing so. The priority will be the timely submission of this paper to JOSS. We can then submit an update to the paper at a later date.

moptis:
Thanks @jordanperr. Re: Anna and Travis as authors, NREL communications will ask us to remove them since they are no longer at NREL. If we want to keep them as authors we'll need up-to-date contact info and I think their approval as well.

jordanperr:
@moptis I have communicated with Anna and Travis, both have provided me with approval and up-to-date contact info.

jordanperr:
read-the-docs check is failing, but it's due to a problem with read-the-docs. We didn't touch anything compiled by them.

https://stackoverflow.com/questions/59846065/read-the-docs-build-fails-with-cannot-import-name-packagefinder-from-pip-in

Issue #69, opened 2020-05-28
Add ERA-5 reanalysis data functionality


Dynorat:
Implement a new reference data set ERA-5 for long term AEP analysis

ERA5 Provider: https://www.ecmwf.int/en/forecasts/datasets/reanalysis-datasets/era5

Add ERA-5 here: https://github.com/NREL/OpenOA/blob/master/operational_analysis/types/reanalysis.py

Dynorat:
Closing this issue due to resolution in pull request being merged.

PR #70, opened 2020-06-03
Adding example data based on open data from ENGIE

ejsimley:
Added a zip file containing original and derived SCADA, meter, curtailment, and reanalysis data for the ENGIE La Haute Borne project to the **examples/data** folder. Created a project class for this project in the **examples** folder.  

The purpose of this new data is to provide a single dataset with SCADA data as well as high temporal resolution meter/curtailment data that can be used in the OpenOA examples.

The location of the data and the license info file can be changed if another folder makes more sense.


nbodini:
Hi all, I've taken a look at the example data to try update the test for the PR #64, and I have a question connected to the ERA-5 data. 
With this PR, ERA-5 has been added to the reanalysis type (as was also asked in issue #69). For how it was written, I think that plant_analysis.py (not just the updated version in my PR#64, but also the version currently in the public repo) requires that data from all the different reanalysis products specified in types/reanalysis.py are available for the wind farm being analyzed (for example in line 427 at https://github.com/NREL/OpenOA/blob/master/operational_analysis/methods/plant_analysis.py). And from the tests I have quickly run, this is why the tests in this PR are failing.

With this in mind, should we further work on plant_analysis.py (and its test) to eliminate this 'requirement' that all reanalysis products are analyzed in the process_reanalysis_data()? One way could be moving the reanal_subset input up from run() to __init__(), so that available reanalysis products for the project are specified right from the beginning. Thoughts?

jordanperr:
Good catch, Nicola. I agree that we need to be able to specify a subset of reanalysis products that would be available for each analysis. I recall that, at some point, it was very important to be able to re-run the AEP analysis with different reanalysis products (without changing any other settings), so this is why we have a renal_subset parameter in run(). Maybe @moptis can clarify if selecting reanalysis products is still needed at runtime?

Two other solutions include:

1. Add reanal_subset to init(), as suggested, but also preserve the reanal_subset argument in run().
2. Modify line 427 such that it only loops over products that are present in the plant. We could even add a helper function to the plant class such as plant.reanalysis_products() which would return non-null products.

moptis:
Yep, good catch Nicola. I agree we don't want to have all possible reanalysis products available for plant_analysis.

Moving reanal_subset up to init makes the most sense to me.

ejsimley:
Thanks for testing this @nbodini - I tried to resolve the issue of being able to specify reanalysis products in my latest commit. The argument reanal_products is now an input to init(), and renal_subset is kept as an input to run(). Let me know how this looks.

nbodini:
Thanks for the additional changes, @ejsimley. I think this looks great now.

I only have another discussion point. In the augmented AEP calculation I've been using to test these data, it is possible to run the calculation using daily average data. If I do so with the ENGIE data, I get one day (2015-02-16) with daily negative energy (-0.000112 GWh), and this raises an exception from the function compute_gross_energy in unit_conversion.py.
Do you think we should modify/remove these negative data in the ENGIE dataset, or change the exceptions in compute_gross_energy, or...?

ejsimley:
I would think that at a daily level (or higher resolution) negative gross energy is realistic, because of the small amount of energy used by turbines when they are not producing power. I would be in favor of removing the requirement that gross energy can't be negative. But I might not be aware of other issues this could cause, so I'm also happy to set a lower limit of zero for meter energy in the Engie project file. @moptis, I'll go with what you think is best here.

ejsimley:
I removed the requirement that gross energy cannot be negative in compute_gross_energy. @nbodini, hopefully this solution works with your code.

Issue #21, opened 2019-06-28
Division by zero error/warning in L5P power curve


sebastianpfaffel:
The L5P power curve returns an error/warning if you query the power value for a wind speed of 0. 

`x = np.linspace(0,20,100)`
....
`plt.plot(x, l5p_curve(x), color="C2", label = 'L5P', linewidth = 3)`

> --> parametric_forms.py:35: RuntimeWarning: divide by zero encountered in power
>   return d + (a - d) / (1 + (x / c) ** b) ** g

jordanperr:
I've encountered this warning before and have spent a little time playing around with it. I was not able to find a robust solution aside from simply silencing the warning, which I decided not to do.

sebastianpfaffel:
Okay, I spend a little time, actually way too much time, looking into this. 

The issue is the "(x / c) ** b" part. b is bound to be negative in the fitting algorithm. Thus we always have a 1/(x/c)**-b situation. When x is 0 it leads to 1/0. 

We have three options:
1. We decide to ignore it and catch the warning. In this case, numpy assumes 1/0 to be 1/0.000......1 and thus inf. In our case this is fine and the result is as close as it gets to what we'd expect it to be.
2. We replace x=0 knowingly by a very small number e.g. 0.00001 using  np.where(x!=0,x,0.000001) and end up with the same result.
`def logistic5param(x, a, b, c, d, g):
    return d + (a - d) / (1 + ( np.where(x!=0,x,0.000001) / c) ** b) ** g `
3. We define the function to return a power of zero at a wind speed of zero. This might lead to unwanted results in case of weird measurements though and is only valid for the purpose of a power curve. We still need to avoid the warning beforehand.
`def logistic5param(x, a, b, c, d, g):
    return np.where(x != 0,
        d + (a - d) / (1 + np.power((x / c), b,where=x!=0)) ** g 
        ,0)`

I'm a little indecisive regarding our options but would tend either to option 1. or option 3. while favoring option 3. It's always nice to start with a 0;0 for a power curve. It's honestly not a very important issue but since we discussed it already I'd like to resolve it. Definitely looks better in the examples.

Any opinions which way to go?

moptis:
@sebastianpfaffel 

Good catch, and thanks for looking into this!

[...]

PR #75, opened 2020-06-19
Fix divide by zero warning in L5P power curve

jordanperr:
Resolves issue #21 by wrapping power curve in an "np.where" clause to catch the case where input power is zero. It returns parameter "d" in the case where x is zero.

Here is the new behavior at x=0 using parameters from the test case:
```
>> from operational_analysis.toolkits.power_curve.parametric_forms import *
>> logistic5param(np.array([1,0.5,0.25,0.1, 0.0]), 1300, -7, 11, 2, 0.5)

array([2.29403585, 2.02598934, 2.00229716, 2.00009298, 2.        ])
```

### DIFF
--- a/operational_analysis/toolkits/power_curve/parametric_forms.py
+++ b/operational_analysis/toolkits/power_curve/parametric_forms.py
@@ -32,7 +32,7 @@ def logistic5param(x, a, b, c, d, g):
         Function[pandas.Series[real]] -> pandas.Series[real]
 
     """
-    return np.where(x!=0, d + (a - d) / (1 + (x / c) ** b) ** g, 0)
+    return np.where(x!=0.0, d + (a - d) / (1 + (x / c) ** b) ** g, d)
 
 
 def logistic5param_capped(x, a, b, c, d, g, lower, upper):
--- a/operational_analysis/toolkits/power_curve/parametric_forms.py
+++ b/operational_analysis/toolkits/power_curve/parametric_forms.py
@@ -35,10 +35,10 @@ def logistic5param(x, a, b, c, d, g):
 
     # In the case where b<0, x==0, there is a divide by zero error. The answer should be "d" when x==0 and b<0.
     if b < 0:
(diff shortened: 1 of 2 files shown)


operational_analysis/toolkits/power_curve/parametric_forms.py
```diff
@@ -32,7 +32,7 @@ def logistic5param(x, a, b, c, d, g):
         Function[pandas.Series[real]] -> pandas.Series[real]
 
     """
-    return d + (a - d) / (1 + (x / c) ** b) ** g
+    return np.where(x!=0.0, d + (a - d) / (1 + (x / c) ** b) ** g, d)
```
ejsimley, line 35: when x == 0, this should return a, not d, since the curve asymptotically approaches a as x approaches 0.

ejsimley: wait, never mind, it approaches a when the slope parameter is positive. and d when negative.

jordanperr: If we restrict b>0 then the issue goes away. This method was originally implemented from the following citation: https://www.sciencedirect.com/science/article/pii/S1364032113007296?via%3Dihub

I checked and it does not seem to place any restriction on the sign of b.

ejsimley: ok, I would suggest something like an if else statement. If b >= 0: use the original line since there is no problem, else: use your new code that fixes the divide by zero issue. 

This way both positive and negative values of b are handled.

ejsimley:
This solves the original issue, but if someone passes x as an array of ints [0,1,2,3,4...], the function returns the logistic curve as integer values. Minor point, but I think it makes more sense to always return an array of floats. What do you think?

jordanperr:
Squash one bug, introduce another!

ejsimley:
Looks good to me now!

PR #77, opened 2020-06-21
Make TIE and electrical losses backward compatible + ENGIE tests

nbodini:
TIE and electrical losses should now be backward compatible, and they should now work with the Examples in PR #71 . 

Also, following @ejsimley 's comments in PR #71 , filter and threshold values (max_power_filter, wind_bin_thresh, correction_threshold) are now not hardcoded anymore in the TIE class.
In the UQ = True case, tuple should be passed for these three parameters (as a range of values is needed in the MC sampling); in the UQ = False case, floats should be passed for these parameters (as no MC sampling is performed on them). I'd need some help on how to make this explicit in the description of the Args, which is used to build documentation. Of course, if you know a better way to express this, please suggest!

### DIFF
--- a/examples/QC_analysis/project_engie_scada.py
+++ b/examples/QC_analysis/project_engie_scada.py
@@ -89,7 +89,7 @@ class Project_Engie(PlantData):
                      "ID"       : "id",
                      "Power_W"              : "wtur_W_avg",
                      "wmet_wdspd_avg"    : "wmet_wdspd_avg", 
-                     "wmet_wDir_avg"    : "wmet_HorWd_Dir"
+                     "wmet_wDir_avg"    : "wmet_wDir_avg"
                      }
 
         self._scada.df.rename(scada_map, axis="columns", inplace=True)
--- a/examples/QC_analysis/project_engie_scada.py
+++ b/examples/QC_analysis/project_engie_scada.py
@@ -67,14 +67,14 @@ class Project_Engie(PlantData):
         #Handle extrema values
         self._scada.df = self._scada.df[(self._scada.df["wmet_wdspd_avg"]>=0.0) & (self._scada.df["wmet_wdspd_avg"]<=40.0)]
         self._scada.df = self._scada.df[(self._scada.df["wtur_W_avg"]>=-1000.0) & (self._scada.df["wtur_W_avg"]<=2200.0)]
-        self._scada.df = self._scada.df[(self._scada.df["wmet_wDir_avg"]>=0.0) & (self._scada.df["wmet_wDir_avg"]<=360.0)]            
+        self._scada.df = self._scada.df[(self._scada.df["wmet_wdir_avg"]>=0.0) & (self._scada.df["wmet_wdir_avg"]<=360.0)]            
 
         logger.info("Flagging unresponsive sensors")
         #Flag repeated values from frozen sensors
         temp_flag = filters.unresponsive_flag(self._scada.df["wmet_wdspd_avg"], 3)
         self._scada.df.loc[temp_flag, 'wmet_wdspd_avg'] = np.nan
-        temp_flag = filters.unresponsive_flag(self._scada.df["wmet_wDir_avg"], 3)
-        self._scada.df.loc[temp_flag, 'wmet_wDir_avg'] = np.nan
+        temp_flag = filters.unresponsive_flag(self._scada.df["wmet_wdir_avg"], 3)
+        self._scada.df.loc[temp_flag, 'wmet_wdir_avg'] = np.nan
         
         # Put power in watts; note although the field name suggests 'watts', it was really reporting in kw
         self._scada.df["Power_W"] = self._scada.df["wtur_W_avg"] * 1000
@@ -89,7 +89,7 @@ class Project_Engie(PlantData):
                      "ID"       : "id",
                      "Power_W"              : "wtur_W_avg",
                      "wmet_wdspd_avg"    : "wmet_wdspd_avg", 
-                     "wmet_wDir_avg"    : "wmet_wDir_avg"
+                     "wmet_wdir_avg"    : "wmet_wdir_avg"
                      }
 
         self._scada.df.rename(scada_map, axis="columns", inplace=True)
--- a/examples/project_ENGIE.py
+++ b/examples/project_ENGIE.py
@@ -124,7 +124,7 @@ class Project_Engie(PlantData):
                     "time"                 : "time",
                     "Wind_turbine_name"    : "id",
                     "Power_W"              : "wtur_W_avg",
-                    "Ws_avg"               : "wmet_HorWdSpd_avg", 
+                    "Ws_avg"               : "wmet_wdspd_avg", 
                     "Wa_avg"               : "wmet_HorWdDir_avg",
                     "Va_avg"               : "wmet_VaneDir_avg", 
                     "Ya_avg"               : "wyaw_YwAng_avg",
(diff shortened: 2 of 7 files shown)


examples/project_ENGIE.py
```diff
@@ -124,7 +124,7 @@ def prepare(self):
                     "time"                 : "time",
                     "Wind_turbine_name"    : "id",
                     "Power_W"              : "wtur_W_avg",
-                    "Ws_avg"               : "wmet_HorWdSpd_avg", 
+                    "Ws_avg"               : "wmet_wdspd_avg",
```
jordanperr, line 127: Shouldn't we switch all of these to lower case, then?

nbodini: Yeah I was thinking the same, I just didn't want to inadvertently break some other codes :)

jordanperr: We use all lower case. Here is the schema file:

https://github.com/NREL/OpenOA/blob/develop/operational_analysis/types/plant_schema_25.json

nbodini: I have updated most of them as lower cases. For now, I have kept wtur_W_avg and wmet_wDir_avg, that are used in this way in some examples and other classes/scripts...

operational_analysis/methods/electrical_losses.py
```diff
@@ -37,29 +37,20 @@ class ElectricalLosses(object):
...
         Initialize electrical losses class with input parameters
         Args:
          plant(:obj:`PlantData object`): PlantData object from which EYAGapAnalysis should draw data.
          num_sim:(:obj:`int`): number of Monte Carlo simulations
-         UQ:(:obj:`bool`): choice whether to perform ('Y') or not ('N') uncertainty quantification
-         uncertainty_meter(:obj:`float`): uncertainty imposed to revenue meter data (for UQ = True case)
-         uncertainty_scada(:obj:`float`): uncertainty imposed to scada data (for UQ = True case)
-         uncertainty_correction_threshold(:obj:`tuple`): The interval of data availability thresholds (fractions) 
-                                                         under which months should be eliminated (for UQ = True case)                                       
+         UQ:(:obj:`bool`): choice whether to perform ('Y') or not ('N') uncertainty quantification
```
ejsimley, line 46: minor update to documentation need for bool

operational_analysis/methods/electrical_losses.py
```diff
@@ -73,17 +64,27 @@ def __init__(self, plant, UQ = False, num_sim = 20000, uncertainty_meter=0.005,
         self._hours_per_day= 24 # Hours per day converter
     
     @logged_method_call
-    def run(self):
+    def run(self, uncertainty_meter=0.005, uncertainty_scada=0.005,
+                 uncertainty_correction_thresh=(0.9,0.995)):
```
ejsimley, line 68: How is the correction threshold specified if the user wants to use UQ = False? In the original version, correction_thresh was an input that defaulted to 0.95. It appears this is hardcoded now on line 127.

operational_analysis/methods/turbine_long_term_gross_energy.py
```diff
@@ -109,14 +94,24 @@ def __init__(self, plant, UQ = False, num_sim = 2000, reanal_subset = ['erai', '
         
 
     @logged_method_call
-    def run(self, enable_plotting = False,
-            plot_dir = None):
+    def run(self,reanal_subset = ['erai', 'ncep2', 'merra2'],
+                 uncertainty_scada=0.005, wind_bin_thresh=(1, 3), 
+                 max_power_filter=(0.8, 0.9), correction_threshold=(0.85, 0.95),
+                 enable_plotting = False,
+                 plot_dir = None):
```
ejsimley, line 101: How are wind_bin_thresh, max_power_filter, and correction_threshold specified is UQ=False? As scalars? Also, please update the docstring to explain the inputs for the UQ and non-UQ cases.

examples/project_ENGIE.py
ejsimley: That's fine with me to stick with the precedent for these variable names. I do think the readability suffers when they're all lowercase though.

operational_analysis/methods/electrical_losses.py
nbodini: Thanks for catching this. I have now changed this.

operational_analysis/methods/turbine_long_term_gross_energy.py
nbodini: Yes, in the UQ = False case, these are single scalar values. I have added this to the docstring description. I am just not sure how to incorporate this in the :obj: part... as we could have both tuples and floats...

ejsimley:
Overall, this looks good to me now. The one thing I would suggest is to check to make sure that input parameters like uncertainty_correction_thresh for electrical losses are tuples when UQ is True and scalars otherwise. Like printing an error message when the input type is not what is expected given the UQ case. Maybe @jordanperr could suggest a way to do this.

nbodini:
I have added a check on the types of the parameters that change between the UQ = True and UQ = False cases in the two classes here. If anyone can suggest a cleaner way to do it, please let me know.

operational_analysis/methods/turbine_long_term_gross_energy.py
```diff
@@ -126,6 +124,30 @@ def run(self, enable_plotting = False,
...
+            if type(correction_threshold) != float:
+                logger.info("correction_threshold needs to be a float if UQ is False!")
+        if self.UQ == True:
+            if type(wind_bin_thresh) != tuple:
+                logger.info("wind_bin_thresh needs to be a tuple if UQ is True!")
+            if type(max_power_filter) != tuple:
+                logger.info("max_power_filter needs to be a tuple if UQ is True!")
+            if type(correction_threshold) != tuple:
+                logger.info("correction_threshold needs to be a tuple if UQ is True!")        
+        # Define relevant uncertainties, to be applied in Monte Carlo sampling
```
jordanperr, line 144: Shouldn't an error be thrown in this case? You can use assert here instead of logger.info:
```
if type(wind_bin_thresh) != float:
    logger.info("wind_bin_thresh needs to be a float if UQ is False!")
```
to
```
assert type(wind_bin_thresh) == float, "wind_bin_thresh needs to be a float if UQ is False!"
```

I don't have a problem with this, even though it's a little repetitive. Input checking code is difficult to shorten without losing clarity. For your own information, If you want to make this shorter, you coule do something like this:
```
vars = [wind_bin_thresh, max_power_filter, correction_threshold]
expected_type = float if self.UQ == True else tuple
if: any([type(x) != expected_type for x in vars]):
    logger.info("...")
```

examples/QC_analysis/project_engie_scada.py
```diff
@@ -89,7 +89,7 @@ def prepare(self):
                      "ID"       : "id",
                      "Power_W"              : "wtur_W_avg",
                      "wmet_wdspd_avg"    : "wmet_wdspd_avg", 
-                     "wmet_wDir_avg"    : "wmet_HorWd_Dir"
+                     "wmet_wDir_avg"    : "wmet_wDir_avg"
```
jordanperr, line 92: Again, why is this capitalized?

nbodini: I've changed this too.

operational_analysis/methods/turbine_long_term_gross_energy.py
nbodini: Cool thanks!

PR #85, opened 2020-07-02
Memoized approach for sample_long_term_reanalysis in plant_analysis

nbodini:
Bringing back the memoized approach for sample_long_term_reanalysis in the plant_analysis class, to speed up the computation.


operational_analysis/methods/plant_analysis.py
```diff
@@ -854,8 +860,33 @@ def sample_long_term_reanalysis(self, n, r):
...
+        Returns:
+           :obj:`pandas.DataFrame`: the IAV-sampled monthly/daily regression inputs
+        """
+    
+        if self.time_resolution == 'D':
+            ws_aggregate_sd_IAV = np.repeat(ws_aggregate_sd_IAV, np.ceil(self.num_days_lt).astype(int))
+        ws_aggregate_sd_IAV.reset_index(drop=True,inplace=True) # TODO: This rectifies a 0-indexing issue. Needs refactoring.
+        iav = np.random.normal(1, ws_aggregate_sd_IAV/long_term_reg_inputs[r])        
+        long_term_reg_inputs_IAV = long_term_reg_inputs.copy()
+        long_term_reg_inputs_IAV[r] = long_term_reg_inputs_IAV[r] * iav
```
jordanperr, line 888: It looks like you're only updating one of the components of long_term_reg_inputs_IAV[r], but returning the whole variable (which has possibly been re-indexed and re-sampled in previous lines)?

operational_analysis/methods/plant_analysis.py
```diff
@@ -825,21 +830,22 @@ def sample_long_term_reanalysis(self, n, r):
            r(:obj:`string`): The reanalysis product used for the current Monte Carlo iteration
         Returns:
            :obj:`pandas.DataFrame`: the windiness-corrected or 'long-term' annualized monthly/daily wind speeds
+           :obj:`pandas.DataFrame`: Long-term wind speed stdev, from which IAV will be derived
         """
+        # Check if valid data has already been calculated and stored. If so, just return it
+        if (r, n) in self.long_term_sampling:
+            long_term_reg_inputs = self.long_term_sampling[(r, n)].copy()
+            ws_aggregate_sd = self.long_term_sampling_sd[(r, n)].copy()
```
jordanperr, line 838: Is there a good reason to perform a copy here? It will increase the memory usage of the program.

operational_analysis/methods/plant_analysis.py
```diff
@@ -854,8 +860,33 @@ def sample_long_term_reanalysis(self, n, r):
...
         # Return result            
-        return long_term_reg_inputs
+        return [long_term_reg_inputs,ws_aggregate_sd]
+
+    @logged_method_call
+    def apply_IAV(self, long_term_reg_inputs, ws_aggregate_sd_IAV, r):
+        """
+        This function calculates IAV for the long-term wind speed data
+        
+        Args:
```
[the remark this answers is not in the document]
nbodini: The function both calculates and applies IAV. I have changed the description to reflect this.

nbodini:
Thanks for the review, Jordan. 
The function apply_IAV calculates and applies IAV to the long-term reanalysis wind speed data. This used to be done within the sample_long_term_reanalysis function, but bringing back the memoized approach broke that. Since IAV has to be randomly sampled, I don't think we can really use both a memoized approach and a randomized IAV sampling within the sample_long_term_reanalysis. So I decided to create a separate function to handle the IAV calculation and application.

Please let me know if this is clear, and if you can think of ways to incorporate the IAV calculation within sample_long_term_reanalysis, as it used to be.

operational_analysis/methods/plant_analysis.py
nbodini: Yes, only the wind speed component of long_term_reg_inputs_IAV is updated (since no IAV is applied to wind direction and temperature, when these are present in long_term_reg_inputs_IAV). Not sure I understood the issue here..

operational_analysis/methods/plant_analysis.py
```diff
@@ -854,8 +860,33 @@ def sample_long_term_reanalysis(self, n, r):
...
+           r(:obj:`string`): The reanalysis product used for the current Monte Carlo iteration
+        Returns:
+           :obj:`pandas.DataFrame`: the IAV-sampled monthly/daily regression inputs
+        """
+    
+        if self.time_resolution == 'D':
+            ws_aggregate_sd_IAV = np.repeat(ws_aggregate_sd_IAV, np.ceil(self.num_days_lt).astype(int))
+        ws_aggregate_sd_IAV.reset_index(drop=True,inplace=True) # TODO: This rectifies a 0-indexing issue. Needs refactoring.
+        iav = np.random.normal(1, ws_aggregate_sd_IAV/long_term_reg_inputs[r])        
+        long_term_reg_inputs_IAV = long_term_reg_inputs.copy()
```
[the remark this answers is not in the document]
nbodini: I agree this copy() is not nice to see. However, without it I noticed some weird behaviors where the wind speed variable after IAV is applied would overwrite the corresponding original dictionary of the memoized approach in sample_long_term_reanalysis. Basically, self.long_term_sampling[(r, n)] would be modified. This is also why I have added the copy() commands in your later comment here... If you know how to avoid this, please let me know!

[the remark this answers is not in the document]
jordanperr: I would prefer to see this copy and the copies on lines 837-838 deleted and then replace lines 839 and 868 with the following:

```
return [long_term_reg_inputs.copy(),ws_aggregate_sd.copy()]
```

This way it is clear when the copy is happening (every time you call sample_long_term_reanalysis)

operational_analysis/methods/plant_analysis.py
jordanperr: Ok. I wonder if there is a way to get rid of the "r" parameter? Didn't we refactor this at some point to self._mc_params? What happened to that?

nbodini: We had created a self._rean_vars variable to save whether temperature and wind components (i.e. wind direction) are chosen by the user. We didn't include wind speed itself in there - probably because there's no option to not use wind speed. 
So for example sample_long_term_reanalysis has a first chunk for wind speed, and then a second chunk for temperature and wind direction (when they are used), controlled by:
namescol = [r + '_' + var for var in self._rean_vars]

operational_analysis/methods/plant_analysis.py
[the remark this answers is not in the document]
nbodini: Looks great to me! Thank you!

operational_analysis/methods/plant_analysis.py
jordanperr: Here - we refactored it to the self._run variable in the turbine_lon_term_gross_energy PR:

https://github.com/NREL/OpenOA/blob/2786cc725e797ef5275083801fee154d0ccf0489/operational_analysis/methods/turbine_long_term_gross_energy.py#L158

And then you can get "r" through self._run.reanalysis_product

https://github.com/NREL/OpenOA/blob/2786cc725e797ef5275083801fee154d0ccf0489/operational_analysis/methods/turbine_long_term_gross_energy.py#L318

nbodini: Yes, that's how both `long_term_reg_inputs_IAV` and `apply_IAV` are called:

```
            # Get long-term regression inputs
            [reg_inputs_lt_noIAV, reg_lt_stdev] = self.sample_long_term_reanalysis(self._run.num_years_windiness,
                                                             self._run.reanalysis_product)

            # Apply IAV to long-term regression inputs
            reg_inputs_lt = self.apply_IAV(reg_inputs_lt_noIAV, reg_lt_stdev, self._run.reanalysis_product)
            
```

Am I missing something?

jordanperr: There is no need to pass in "self._" anything to these methods because you already have access to "self" automatically. 

Can't we just use self._run.reanalysis_product self. _run.num_years_windiness from within self.sample_long_term_reanalysis and self.apply_IAV, getting rid of both arguments to sample_long_term_reanalysis and and getting rid of the 3rd argument to apply_IAV?

nbodini: Ahh, got it! I've now changed this.

jordanperr:
I fixed a conflict in the docstring using Github's UI, hence the cryptic merge commit.

PR #89, opened 2020-07-29
Delete old examples

jordanperr:
Removing the old examples from the examples folder.


jordanperr:
merging this even though rtd fails because this should get fixed by Rob's PR 88, which I will merge momentarily.

PR #90, opened 2020-08-05
Pip packaging

jordanperr:
- Auth token for Pypi has been added as PYPI_PASSWORD secret.
- Pypi packaging action is added to .github workflows
- Documentation updated to include instructions on how to tag for pip package
- Version number bump to 2.0
- Removed pytestrunner from setup.py (we don't use it anymore)
- Added read_file function to setup.py which is used to read README.md as the pypi description.

A sample of what this looks like on PyPi is can be seen here: https://test.pypi.org/project/OpenOA/
The pypi url is: https:/pypi.org/project/OpenOA/ 
Happy to add folks as maintainers of the PyPi project (we should probably have at least two maintainers).

### DIFF
--- a/.github/workflows/tags-to-pypi.yml
+++ b/.github/workflows/tags-to-pypi.yml
@@ -1,4 +1,4 @@
-name: Publish any new version tags (matching 'v*') to PyPi as a new release.
+name: PyPi Publisher
 
 on:
     push:
@@ -7,7 +7,7 @@ on:
 
 jobs:
     build-n-publish:
-        name: Build and publish Python 🐍 distributions 📦 to PyPI and TestPyPI
+        name: Build and publish Python package to PyPi
         runs-on: ubuntu-18.04
         steps:
             - uses: actions/checkout@master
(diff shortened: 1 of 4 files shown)


.github/workflows/tags-to-pypi.yml
```diff
@@ -0,0 +1,27 @@
+name: PyPi Publisher
```
RHammond2, line 1: I haven't seen them set up this way and based the `floris` [version](https://github.com/NREL/floris/blob/master/.github/workflows/python-publish.yml) on the GH Actions [example](https://docs.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#publishing-to-package-registries). I noticed that you're using tags instead of releases which is likely why it is different so I'll trust the test version transfers over to the real world correctly and that your know-how are correct here.

jordanperr: I merely followed a different guide to set this up: https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/

I wonder why the guides are so different between python.org and github.com?

I think it's nice to use tags because they're easy to create (and test). I believe github releases are based on tags.

RHammond2: That's fair enough. I assume it would come down to the workflows for whoever created (and approved) the examples.¯\_(ツ)_/¯

PR #97, opened 2020-08-25
adding authorship

RHammond2:
Starting a PR to reconsider the JOSS paper author order (per discussion with Jordan) as well as adding my own name now that I am a part of the project.

I'd recommend also linking to a GitHub/ORCID/Google Scholar type page for the different authors in the `readme.md`.


nbodini:
Sorry if I jump on this, but I think Jordan had agreed a while ago to add my name to the list of the co-authors of the paper, too. Should I add it here, or...?

PR #101, opened 2020-09-16
Temporarily fix gdal dependency by removing GeoPandas requirement

ejsimley:
Removed the GeoPandas installation requirement from setup.py and moved the GeoPandas import statement in the `AssetData` class to the `AssetData.parse_geometry` method where GeoPandas is used. This should resolve the GDAL dependency issue many Windows users are having for now, since `AssetData` is currently not being used.


ejsimley:
Feel free to update documentation as you see fit.

readme.md
```diff
@@ -33,7 +33,7 @@ conda activate openoa-env
 
 #### Special Note for users of Microsoft Windows:
 
-The Anaconda python distribution is *required* for users of Microsoft Windows. This is because the pip package of GDAL for Windows requires Visual Studio to compile some of the dependencies. While advanced users are welcome to explore this option, we find it is easier to install the following packages via Anaconda:
+If using the 'AssetData' class, the Anaconda python distribution is *required* for users of Microsoft Windows. This is because of the GDAL package used in this class. The pip package of GDAL for Windows requires Visual Studio to compile some of the dependencies. While advanced users are welcome to explore this option, we find it is easier to install the following packages via Anaconda:
```
RHammond2, line 36: I might change single quotes to using back quotes (the other option on the tilde button) around Asset Data, so it displays in monospace like this: `AssetData`, compared to quoted like this: 'AssetData'. That's more stylistic, but helps to signal that this is a piece of the code.

operational_analysis/types/asset.py
```diff
@@ -11,7 +10,12 @@ class AssetData(object):
     """
     This class wraps around a GeoPandas dataframe that contains
     metadata about the plant assets. It provides some useful functions
-    to work with this data (e.g., calculating nearest neighbors, etc.)
+    to work with this data (e.g., calculating nearest neighbors, etc.). 
+    
+    This class requires the geopandas package, which can be installed using:
```
RHammond2, line 15: I would recommend making this a "Note" like in the sphinx example so that when someone reads the docstrings it is further highlighted (example below).
image

RHammond2 (approved):
I left a couple of pretty minor suggestions in terms of documentation styling to help emphasize some pieces, but they're not required, just nice to have. Either way, I think this is good.

operational_analysis/types/asset.py
ejsimley: Good suggestion, it looks great in the documentation.

jordanperr (approved):

jordanperr (changes requested):
Need to fix readthedocs error. Looks like it's an error on RTD side:

https://github.com/sphinx-doc/sphinx/issues/8198
https://github.com/readthedocs/readthedocs.org/issues/7492

ejsimley:
Going to merge despite the readthedocs issue, which will hopefully get resolved soon.

Issue #99, opened 2020-09-01
Revisit long-term correction in AEP class


nbodini:
This issue is to suggest a modification of the long-term (LT) correction applied in the AEP class. 

With the current LT approach, the POR regression is applied to a single average year that represents the long-term conditions. So at monthly resolution, the POR regression is applied to a single set of 12 data points, obtained as the long-term average January, February, and so on. At daily resolution, the POR regression is applied to a single set of 366 data points, obtained as the long-term average January 1, January 2, and so on.
When doing so, we are NOT preserving the full variability of the long-term inputs (wind speed, temperature and wind direction when used). In this figure, I compared the distribution of wind speed during the period of record (in blue) with the distribution of a single average year calculated from the 20-year LT record:

![Picture12](https://user-images.githubusercontent.com/55894604/93916027-21433c00-fcc6-11ea-9053-222ea5c07ecb.png)

It is clear that the current LT correction approach suppresses most of the variability in (in this case) wind speed. When a linear regression is used, this difference in distributions doesn't have an impact on the overall mean AEP. However, when _nonlinear_ ML models are used, this will have an impact on the mean AEP predicted in the LT correction. 

A new proposed LT approach would apply the POR regression to the actual full long-term period. So at monthly resolution with let's say 20 years used for the LT correction, the POR regression would be applied to 12x20 data points. At daily resolution with 20 years used for the LT correction, the POR regression would be applied to 365.25x20 data points.
In this way, we would be preserving the variability of the data:

![Picture1](https://user-images.githubusercontent.com/55894604/93915533-6b77ed80-fcc5-11ea-92a9-75dc8a6bdfb9.png)

One potential issue connected to the application of the new proposed LT approach is that the LT period might have data which are outside of the training range of the ML models. To overcome this issue, we need to set an upper filter/limit to the data. We suggest this to be set to the maximum value observed in the period of record, but are happy to hear about other suggestions as we realize this represents some sort of arbitrary choice. 

As a quick solution, I suggest incorporating the proposed new LT correction in the AEP class, and temporarily turn off IAV while we wait for a more structured solution to incorporate IAV.

nbodini:
This issue is being addressed in PR #102

Issue #86, opened 2020-07-09
Revisit IAV in AEP class


nbodini:
This issue is to discuss a couple of things that we may need to change connected to how inter-annual variability is calculated and applied in the AEP class.

First, here's a quick reminder on IAV is currently calculated and applied within the Monte Carlo framework:
When AEP is calculated at monthly resolution:
- Calculate LT monthly average wind speed;
- Group data by calendar month;
- Calculate 12 values of the Coefficient of Variation;
- Apply IAV to LT monthly average wind speed by sampling from a normal distribution with mean 1 and assigned CoV for the corresponding month.

When AEP is calculated at daily resolution:
- Calculate LT daily average wind speed;
- Group data by calendar month;
- Calculate 12 values of the Coefficient of Variation using all the daily average values in each calendar month;
- Apply IAV to LT daily average wind speed by sampling from a normal distribution with mean 1 and assigned CoV for the corresponding month.

The first issue here is that the IAV contribution to AEP uncertainty is much lower when the calculation is performed at daily resolution. We suspect this is due to the high autocorrelation in the data at daily resolution, which is currently not taken into account in the way IAV is calculated.

The second issue is connected to how IAV works when applied to the new LT approach (proposed in issue #99 ) and a nonlinear ML regression model. When IAV is applied to the new LT correction with a nonlinear ML model, the resulting mean AEP is generally much higher (up to +10%) than what it should be.
At each MC iteration, IAV makes the wind speed distribution wider in a symmetric way. However, this determines an unrealistic response when a non-symmetrical ML regression is used to go from wind speed to power. See the example below:

![Screen Shot 2020-09-22 at 12 26 44 PM](https://user-images.githubusercontent.com/55894604/93922240-e72a6800-fcce-11ea-98e3-c515b33211f8.png)

The blue distribution is the original wind speed distribution (i.e. without IAV applied) from a 20-year record. The orange distribution shows the symmetric widening of the distribution when IAV is applied. The bottom plot shows the impact that this has when a nonlinear model is used. The daily power from the two ‘more of this’ regions does NOT compensate: there’s a net surplus of power when IAV is applied to a non-linear model.

To fix these issues, we suggest to move IAV out of the Monte Carlo framework, so that IAV is applied after the regression model is applied to the long-term period.

This would likely require quite a bit of work, and help would be much appreciated.

PR #102, opened 2020-09-29
AEP: update long-term correction and remove IAV for now

nbodini:
This PR addresses the issue described in #99 
It also adds/fixes a couple minor things:
- "era5" is now one of the default reanal_products in __init__
- mse and r2 are now saved in sim_results

Note that IAV has been temporarily turned off because of the issues described in #86


RHammond2 (changes requested):
From a software perspective, this looks good, and I'd just update the one docstring so that it matches what is being returned with this update.

operational_analysis/methods/plant_analysis.py
```diff
@@ -873,33 +867,25 @@ def sample_long_term_reanalysis(self):
         
         Args:
            (None)
-
         Returns:
            :obj:`pandas.DataFrame`: the windiness-corrected or 'long-term' annualized monthly/daily wind speeds
            :obj:`pandas.DataFrame`: Long-term wind speed stdev, from which IAV will be derived
```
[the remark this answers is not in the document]
nbodini: Good catch!

operational_analysis/methods/plant_analysis.py
```diff
@@ -844,14 +835,14 @@ def run_AEP_monte_carlo(self):
             gross_por = fitted_model.predict(np.array(pd.concat(reg_inputs_por, axis = 1)))
 
             if self.time_resolution == 'M': # Undo normalization to 30-day months
-                gross_lt = gross_lt*self.num_days_lt/30 
+                gross_lt = gross_lt*np.tile(self.num_days_lt,self._run.num_years_windiness)/30 
                 gross_por = gross_por*self.num_days_lt/30
```
ejsimley, line 839: As we were discussing in PR #105 we should update how gross_por is calculated, so it uses the new LT correction method too, and uses inputs from _reanalysis_por instead of _reanalysis_por_avg.

ejsimley (commented):
The new LT correction method, and removing IAV for now, looks good to me. As we discussed, we still need to update how gross_por is calculated. I was debating if that should be a separate issue/PR, but I think we could just resolve that here. @nbodini, feel free to try to fix that, but I'm happy to take a look at it too. Let me know what you think!

operational_analysis/methods/plant_analysis.py
nbodini: @ejsimley I agree it makes sense to fix this issue here. I've taken a first look and it's a bit more complicated than what I was expecting, because we need to be careful about seasonality impacts: if we calculate gross_por simply using _reanalysis_por, we won't necessarily get a gross energy estimate that is equally distributed throughout the year (e.g. if the POR goes from December 2016 to February 2020, winter would be weighted more in gross_por, and this should be taken into account when then calculating lt_por_ratio). Let me know if you can think of an easy way to solve this.
 
In the meantime, I've noticed an issue in how lt_por_ratio was being calculated, and I've fixed it.

RHammond2 (approved):
From my standpoint, this looks good with the updated docstrings.

RHammond2:
@jordanperr have you seen an error like this before on the RTD build? The issue seems to be stemming from the usage of easy_install and building from an egg rather than wheel, but in the binder PR I don't have this issue and it was pre-3.0, so maybe that's another part of it. I haven't been able to find much on this issue other than [eggs are deprecated and we're supposed to use wheels](https://setuptools.readthedocs.io/en/latest/userguide/commands.html#bdist-egg-create-a-python-egg-for-the-project). Happy to keep digging if you don't have and leads on this.

```
Installed /home/docs/checkouts/readthedocs.org/user_builds/openoa/envs/102/lib/python3.7/site-packages/Shapely-1.7.1-py3.7-linux-x86_64.egg
Searching for pyproj
Reading https://pypi.org/simple/pyproj/
Downloading https://files.pythonhosted.org/packages/17/e5/3f5cdff3e955bcd768cdb0f4236f2d6e022aaa72f57caf7f4d5f552c88fc/pyproj-3.0.0.post1.tar.gz#sha256=a49581629cadd29e61fc061d153a4d62ff28b4063c71fe8ca881eeb98cd22017
Best match: pyproj 3.0.0.post1
Processing pyproj-3.0.0.post1.tar.gz
Writing /tmp/easy_install-frr172od/pyproj-3.0.0.post1/setup.cfg
Running pyproj-3.0.0.post1/setup.py -q bdist_egg --dist-dir /tmp/easy_install-frr172od/pyproj-3.0.0.post1/egg-dist-tmp-eww3ihtg
error: Setup script exited with ERROR: Cython.Build.cythonize not found. Cython is required to build from a repo.
```

nbodini:
I have now added a couple of other minor edits still connected to the LT correction:
1) leap years are now better accounted for in the sample_long_term_reanalysis function, when the AEP calculation is performed at daily resolution. For this, self._calendar_samples at daily resolution has been now set to 365.25 (instead of 365).
2) I have fixed the issue (noted by Jordan in the comments of the class) of some abuse of the reset_index within the sample_long_term_reanalysis function. Now the output of that function keeps as index the time information (while before this info was being lost).

operational_analysis/methods/plant_analysis.py
nbodini: @ejsimley, I've been thinking a bit more about this, and still couldn't find an easy fix. I guess one option could be trimming the POR data and calculating gross_por with the new LT approach using n*12 months. Using the old LT correction to calculate gross_por doesn't make much sense to me (and would lead to misleading results when using nonlinear ML models). The other option would be to just remove from the class the couple of lines to calculate gross_por (it's just used to calculate lt_por_ratio which is given as output of the class, not sure how important that is?), but that's not good either...

nbodini: @ejsimley I think I was able to fix this! gross_por is now calculated using the new LT approach (i.e., the regression model is applied to the raw por data, without calculating any intermediate averages), and I've deleted _reanalysis_por_avg as it is not needed anymore with the new approach. To fix the seasonality issue described above, self.groupby_time_res is applied to gross_por to calculate an average gross energy value for each month (so 12 values total), and this "average" year is then used to calculate lt_por_ratio.

ejsimley: I agree this incorporates the new LT approach while avoiding seasonality issues. Looks good to me!

ejsimley (approved):
The latest changes look good and the examples are all running fine. OK with me to merge after updating the documentation where I left a comment. Perhaps Jordan wants to look at the changes too.

operational_analysis/methods/plant_analysis.py
```diff
@@ -873,33 +867,21 @@ def sample_long_term_reanalysis(self):
         
         Args:
            (None)
-
         Returns:
            :obj:`pandas.DataFrame`: the windiness-corrected or 'long-term' annualized monthly/daily wind speeds
```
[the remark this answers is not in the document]
nbodini: I've updated the documentation!

ejsimley:
@jordanperr Note that I created issue #123 related to AEP analysis failing when wind direction is added as a regression variable. But since the code fails in `apply_IAV` this PR (which removes IAV for now) should resolve the issue.

jordanperr:
Looks good to me. Thanks for your contribution, Nicola! I apologize for the lengthly review period.

PR #104, opened 2020-09-29
Upgrade all toolkits to take Series instead of Dataframes

### LINKED ISSUE #12: Convert all toolkit functions to operate on Pandas.Series and Numpy.Array
Some functions in toolkits take a dataframe as their first argument, and then strings that represent column names of this dataframe. We'd like to migrate these functions to operate on Series and Array, avoiding the need to pass the entire dataframe and string column names. This will allow for more flexibility in their use. The following functions shall be converted:

- [x] met_toolkit.py ( merge in PR #49 )

- compute_air_density
- air_density_adjusted_wind_speed
- compute_turbulence_intensity
- compute_shear
- compute_veer

- [x] imputing.py

- correlation_matrix_by_id_column
- impute_data
- impute_all_assets_by_correlation

- [x] unit_conversion.py

- convert_feet_to_meter (documentation)
- convert_power_to_energy (documentation)

RHammond2:
This pull request closes #12 and is ready to merge ~~after a decision is made with the one failing test~~.

I've updated all the instances where imputation is performed to align with the revised API for the imputation methods. There is a slight difference in some calculations in these methods so that they can take array-like objects (aka `np.ndarray` and `pd.Series`), but overall I kept changes to a minimum.

~~There is one part of a test that is failing, screenshot below, likely due to a rounding error from differences between `np.int64` and `int`.  @jordanperr one other option that could be used here is to use [`numpy.testing.assert_allclose`](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html#numpy.testing.assert_allclose) for testing within a tolerance level as opposed to with matching out to a certain number of decimals.~~

image


RHammond2:
I've updated the imputation methods to convert array/Series objects to floats when returning, so all tests now pass.

operational_analysis/toolkits/imputing.py
```diff
@@ -7,42 +7,44 @@
...
 
     Returns:
         :obj:`pandas.DataFrame`: Correlation matrix with <id_col> as index and column names
     """
-    assets = df[id_col].unique()  # List different assets in <id_col>
+    assets = np.unique(id_col)  # List different assets in <id_col>
     corr_df = pd.DataFrame(index=assets, columns=assets, dtype=float,
                            data=np.nan)  # Define correlation matrix that is set to NaN by default
+    df = pd.DataFrame(np.column_stack((id_col, align_col, value_col)), columns=["id_col", "align_col", "value_col"])
+    df.value_col = df.value_col.astype(float)
```
jordanperr, line 25: For example, is this going to incur a performance penalty in the case where id_col, align_col, and value_col are coming from a separate pandas datafame?

operational_analysis/methods/turbine_long_term_gross_energy.py
```diff
@@ -381,11 +381,13 @@ def filter_sum_impute_scada(self):
...
-                                                                                              ref_col = 'energy_kwh_corr',
-                                                                                              align_col = 'day',
-                                                                                              id_col = 'id')
+        self._scada_daily_valid['energy_imputed'] = imputing.impute_all_assets_by_correlation(
+            input_col=self._scada_daily_valid.energy_kwh_corr,
+            ref_col=self._scada_daily_valid.energy_kwh_corr,
+            align_col=self._scada_daily_valid.day,
+            id_col=self._scada_daily_valid.id,
+            index_col=self._scada_daily_valid.index.values,
+        )
```
jordanperr, line 390: _scada_daily_valid is already a pandas dataframe.

operational_analysis/toolkits/imputing.py
RHammond2: I would assume there is a performance hit if they are all coming from the same data `pd.DataFrame`, because we're effectively just putting the data back together again compared to three separate arrays. I assumed the intent was to make so `np.ndarray` and `pd.Series` objects could be passed into the function, or is there something I was missing?

operational_analysis/methods/turbine_long_term_gross_energy.py
RHammond2: Yeah, that's my understanding as well, this was to get the functionality in line with what's expected for input types for the imputing function call.

RHammond2:
@jordanperr  would it help to have a wrapper that accepts the data frames and column names in addition to the array objects? That wouldn't be much work to add in.

RHammond2:
@jordanperr here are some results for you based off using the tests for timing:

In the below screenshots, T1 and T2 refer to the two tests in the `test_imputing_toolkit.test_correlation_matrix_by_id_column()` test, [here](https://github.com/NREL/OpenOA/blob/enhancement/issue_12/test/unit/test_imputing_toolkit.py#L94-L96) and [here](https://github.com/NREL/OpenOA/blob/enhancement/issue_12/test/unit/test_imputing_toolkit.py#L98-L100), respectively.

I computed the mean and 95% confidence intervals on the timing using 1000 runs for each, with all timings reported in milliseconds. The new method is slightly slower, but not by a huge amount.

Original implementation (v1):
image


pd.Series values passed (v2):
image


np.ndarray values passed (v3):
image


Let me know if there is a more comprehensive set of timing analyses to run or if you'd like to see the code I used for this! I'm also happy to reimplement the method in a potentially more sped up fashion rather than simply enabling a new interface if that would help.

RHammond2:
v3 is essentially doing all of this work, but better, so this is getting closed permanently.

PR #108, opened 2020-10-12
AEP: fast track ML hyperparameter selection

nbodini:
This PR changes the way ML hyperparameters are chosen in the plant_analysis class. 
In the old version, hyperparameters were optimized at each Monte Carlo iteration, which was causing the ML models to be very slow to run (over a minute per iteration for some of them). 
With this PR, we introduce a memoized approach where the hyperparameters are optimized just once for each reanalysis products, and then stored in a dictionary. This significantly speeds up the computation.


operational_analysis/methods/plant_analysis.py
```diff
@@ -776,12 +777,24 @@ def run_regression(self, n):
         # Machine learning models
         else: 
             ml = MachineLearningSetup(self.reg_model, **self.ml_setup_kwargs)
-            ml.hyper_optimize(np.array(reg_data[:,0:-1]), reg_data[:,-1], n_iter_search = 5, report = False, cv = KFold(n_splits = 2))
-            predicted_y = ml.random_search.predict(np.array(reg_data[:,0:-1]))
+            # Memoized approach for optimized hyperparameters
+            if self._run.reanalysis_product in self.opt_model:
+                self.opt_model[(self._run.reanalysis_product)].fit(np.array(reg_data[:,0:-1]), reg_data[:,-1])
+                predicted_y = self.opt_model[(self._run.reanalysis_product)].predict(np.array(reg_data[:,0:-1]))
```
ejsimley, line 783: Here `fit` and `predict` are being called from self.opt_model, which is defined in `machine_learning_setup` as `self.opt_model = self.random_search.best_estimator_`. But below after the `else` statement, `predict` is called from `ml.random_search`. Also, within machine_learning_setup the model is fit using `self.random_search.fit`. Why are `fit` and `predict` being called using `ml.random_search.best_estimator_.fit` in one case and `ml.random_search.fit` in the other? Is there a reason for saving the opt_model as `self.random_search.best_estimator_` instead of just `self.random_search` in machine_learning_setup?

ejsimley (approved):
This looks fine to me, although I did post a question. More generally, how did you find that the hyperparameters only vary depending on reanalysis product, but not the specific random sample of data points in the MC iteration? Otherwise, looks like this should result in a significant speedup!

nbodini:
Thanks for reviewing this, @ejsimley ! 
To answer your general question, Mike and I thought that since the hyperparameter selection process already involves cross-validation using an 80-20 split, the selected combination of hyperparameters should be robust regardless of the bootstrapping done in the MC iterations. Basically, we're already re-sampling the data to determine the "best" hyperparameters, so the MC bootstrapping shouldn't impact that too much. 
Of course, this is not true anymore when considering different reanalysis products, since they have fundamentally different data, and that's why we decided to keep one set of hyperparameters for each product. Makes sense?

operational_analysis/methods/plant_analysis.py
nbodini: Thanks for catching this inconsistency. 
I've re-factored this chunk of the code, so that now predict is always called using the `best_estimator_`.
In machine_learning_setup, it is correct to call fit on `ml.random_search`, as `best_estimator_` becomes available only after `ml.random_search` is fitted first.
Basically, the idea is: fit random_search only once for each reanalysis product, and then use its `best_estimator_` for all the rest of the calculations.

ejsimley: Got it, that makes more sense and the code looks a lot cleaner, especially removing some duplication. One last question, when the `opt_model` is already found in the memoized dictionary, you are still calling `fit` on the saved `best_estimator_`. Is that step necessary? Does `ml.random_search.fit` set the hyperparameters and model parameters, and `ml.random_search.best_estimator_` just set the model parameters for the specific data sample? Just trying to understand where the time savings comes from if `fit` is still called.

ejsimley:
Thanks @nbodini, yes that makes sense! The 80/20 cross validation in the hyper parameter selection explains why this is needed only once.

operational_analysis/methods/plant_analysis.py
nbodini: Yes, the additional fit is needed to set the model weights to the specific data used at each MC iteration, but not the hyperparameters, which are taken from the `opt_model`. So the time savings come from skipping the cross validation needed to set the hyperparameters, so that only the fitting of the model weights is performed at each iteration.

operational_analysis/methods/plant_analysis.py
```diff
@@ -774,15 +775,22 @@ def run_regression(self, n):
...
             return reg
         # Machine learning models
-        else: 
+        else:
             ml = MachineLearningSetup(self.reg_model, **self.ml_setup_kwargs)
-            ml.hyper_optimize(np.array(reg_data[:,0:-1]), reg_data[:,-1], n_iter_search = 5, report = False, cv = KFold(n_splits = 2))
-            predicted_y = ml.random_search.predict(np.array(reg_data[:,0:-1]))
+            # Memoized approach for optimized hyperparameters
+            if self._run.reanalysis_product in self.opt_model:
+                self.opt_model[(self._run.reanalysis_product)].fit(np.array(reg_data[:,0:-1]), reg_data[:,-1])
```
jordanperr, line 782: Would be useful to add a comment saying the last column of reg_data is the target variable for regression. This was a little confusing to me re-reading the code.

operational_analysis/methods/plant_analysis.py
```diff
@@ -774,15 +775,22 @@ def run_regression(self, n):
...
+            else: # optimize hyperparameters once for each reanalysis product  
+                ml.hyper_optimize(np.array(reg_data[:,0:-1]), reg_data[:,-1], n_iter_search = 20, report = False, cv = KFold(n_splits = 5))
+                # Store optimized hyperparameters for each reanalysis product
+                self.opt_model[(self._run.reanalysis_product)] = ml.opt_model
```
jordanperr, line 786: Do we also need to run line 782 again, here? I don't see any call to opt_model.fit(). Furthermore, I don't see any member MachineLearningSetup.opt_model in the source. Where is that variable coming from, and when is it fit if line 782 has not yet been run?

jordanperr:
Hi Nicola, It looks like there is a conflict now with plant_analysis.py. I also made two inline comments in your code. If you have the time to fix the conflict and review my inline comments, that could speed up the review. Otherwise, I will plan to revisit this later in the week.

operational_analysis/methods/plant_analysis.py
nbodini: Hi Jordan,  MachineLearningSetup.opt_model is assigned in machine_learning_setup.py, line 147. It is the result of the fit in line 143 there, so that's why the fit is not performed again in plant_analysis.py. On the other hand, we need to re-perform the fit in line 782 since in that case we're using a different set of input data to the regression. Hope this is clear enough, let me know if not!

operational_analysis/methods/plant_analysis.py
nbodini: I've now added a comment!

nbodini:
Hi @jordanperr , I've fixed the conflict and answered to your inline comments. Let me know if my answers don't make sense!

operational_analysis/methods/plant_analysis.py
jordanperr: I see. I was looking at the develop branch and missed your addition to that file in this PR. This looks good. Your comments make sense and resolve the issue.

jordanperr (approved):

PR #109, opened 2020-10-13
Merge develop into master - v2.0.1

RHammond2:
@jordanperr I don't think I have the proper access to merge this into the master branch. Would you mind updating privileges or merging when you get a chance?


operational_analysis/types/asset.py
```diff
@@ -124,11 +128,11 @@ def calculate_nearest(self, active_turbine_ids, active_tower_ids):
 
     def distance_matrix(self):
         ret = np.ones((self._asset.shape[0], self._asset.shape[0])) * -1
-        for i, j in itertools.permutations(self._asset.index, 2):
+        for i, j in itertools.combinations(self._asset.index, 2):
             point1 = self._asset.loc[i, 'geometry']
             point2 = self._asset.loc[j, 'geometry']
             distance = point1.distance(point2)
-            ret[i][j] = distance
+            ret[i, j] = ret[j, i] = distance
```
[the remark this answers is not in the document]
RHammond2: Exactly correct!

operational_analysis/types/asset.py
```diff
@@ -1,17 +1,23 @@
 import importlib
 import itertools
 
-import geopandas as gp
 import numpy as np
 import pandas as pd
+from pyproj import Transformer
```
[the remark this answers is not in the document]
RHammond2: I don't believe that GDAL is used for pyrpoj or shapely, so this should be ok. In any case in `AssetData.parse_geometry()` there is a coordinate reference system conversion and that has to be handled with some geographic library. If the idea is to remove that functionality altogether, then I think we could simply use `geopy` for their geodesic distance function for "correct" turbine distances, or we could do a simple haversine distance implementation.

operational_analysis/types/asset.py
```diff
@@ -1,17 +1,23 @@
...
     This class wraps around a GeoPandas dataframe that contains
     metadata about the plant assets. It provides some useful functions
-    to work with this data (e.g., calculating nearest neighbors, etc.)
+    to work with this data (e.g., calculating nearest neighbors, etc.). 
+    
+    Note:
+        This class requires the geopandas package, which can be installed using:
+            pip install geopandas
+
+        If using Microsoft Windows, please see the 'Requirements' section of the OpenOA readme file for instructions on installing geopandas.
```
ejsimley, line 20: @RHammond2 great that you found a good alternative to geopandas. I added this note because of the earlier temporary fix where we were still using geopandas but not making it a requirement. So I would remove these lines since they aren't relevant anymore.

readme.md
```diff
@@ -33,7 +33,7 @@ conda activate openoa-env
 
 #### Special Note for users of Microsoft Windows:
 
-The Anaconda python distribution is *required* for users of Microsoft Windows. This is because the pip package of GDAL for Windows requires Visual Studio to compile some of the dependencies. While advanced users are welcome to explore this option, we find it is easier to install the following packages via Anaconda:
+If using the `AssetData` class, the Anaconda python distribution is *required* for users of Microsoft Windows. This is because of the GDAL package used in this class. The pip package of GDAL for Windows requires Visual Studio to compile some of the dependencies. While advanced users are welcome to explore this option, we find it is easier to install the following packages via Anaconda:
```
ejsimley, line 36: Similarly, this note about the Anaconda requirement for Windows and the GDAL package isn't relevant anymore, so I believe should be removed.

operational_analysis/types/asset.py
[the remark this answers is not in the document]
jordanperr: Eric and I spoke, sounds good!

CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 # Changelog
 All notable changes to this project will be documented in this file. If you make a notable change to the project, please add a line describing the change to the "unreleased" section. The maintainers will make an effort to keep the [Github Releases](https://github.com/NREL/OpenOA/releases) page up to date with this changelog. The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## [2.0.1 - 2020-10-13]
+- Replaced `GeoPandas` functionality with `pyproj` and `Shapely` for coordinate
+reference system conversion and distance measurements.
+
```
jordanperr, line 7: Please add "Moved and renamed tests and updated the documentation accordingly."

jordanperr (approved):
Pending the small changes already mentioned here, I think it's good to go. Please wait for @ejsimley 's approval before merging.

operational_analysis/types/asset.py
RHammond2: Good catch! I'll update the doctsring.

ejsimley (approved):
This looks good to me now. Feel free to merge! Only thing I would say is that Rob can add his name to the contributors list in the readme.

CHANGELOG.md
RHammond2: Done.

RHammond2:
@ejsimley my name was added in a different PR, but I've just gone and added here as well. Should I merge or wait for other changes to be merged into develop?

ejsimley:
@RHammond2 I suggest merging this in now to take care of the dependency issues some users are having.

PR #111, opened 2020-11-02
AEP: update bootstrapping approach

nbodini:
This PR fixes how the bootstrapping approach to quantify regression uncertainty is applied in the AEP class.

Previously, a random 80% of the POR data was selected, at each Monte Carlo iteration, as input to the regression model. However, this is now how a proper bootstrapping work, and this poses some uncertainty connected to the choice of the 80% (i.e. how does uncertainty vary if we select 60% instead, or 90%?).
With this updated version, a proper bootstrapping approach is used, where data are sampled with replacement, and the sample size is 100% of the POR data.


ejsimley (approved):
This is a good change to the bootstrapping method!

PR #112, opened 2020-11-10
updating for pyproj oddity on rtd build

RHammond2:
Testing to figure out how to solve this pyproj installation issue on the Read The Docs site.


jordanperr:
Fixed through rtd admin interface.

Issue #114, opened 2020-12-11
Install fails with Python 3.9


gschivley:
The installation instructions say to do

```
conda create --name openoa-env python=3
```

which now installs python 3.9. I tried to install one final time (previous laptop died) to finish my review, and it looks like there are issues (maybe some missing wheels, maybe just on Mac?). I had to specify python 3.8 to install.

Part of review (openjournals/joss-reviews#2171)

brynpickering:
I also came across this issue on OSX, similarly solved by forcing `python=3.8`.

Note: this comment comes to you as part of my wider [JOSS review](https://github.com/openjournals/joss-reviews/issues/2171)

jordanperr:
This seems to be an issue with many packages right now. Please proceed reviewing OpenOA assuming the Python 3.8 dependency - and we will work to figure out the long term solution for Python 3.9.

jordanperr:
The main reason for install failure is that Shapely does not have binary wheels for Mac + Python 3.9. This is an issue specific to Macs running Python 3.9. There is an open ticket for this in the Shapely issue tracker. People in this thread suggest using brew to first install geos (which is the critical missing dependency for Shapely) before trying to install Shapely. This workaround will likely work for OpenOA, but I am have not tested it.

For now:
- Readme will be updated to recommend Python 3.8.
- I will leave this issue open until OpenOA supports Python 3.9 on all platforms.

Thanks again for reporting this.

https://github.com/Toblerity/Shapely/issues/1040

PR #117, opened 2020-12-15
Bug Fix: missing requirements

### LINKED ISSUE #115: Example notebooks package requirements
It would be nice to jump into the example notebooks without having to deal with `ModuleNotFound` errors once working in them. Would it be possible to have a requirements file in the examples directory, containing all the additional packages required in the example notebooks? It could also include jupyter itself (or, at least, ipykernel). 

Note: this issue comes to you as part of my wider [JOSS review](https://github.com/openjournals/joss-reviews/issues/2171)

### LINKED ISSUE #116: Missing h5pyd package requirement
The required dependency `h5pyd` is missing following a fresh install of OpenOA. 

This causes an error in the file `operational_analysis/methods/quality_check_automation.py`. This error is not caught by tests since the file coverage is 0%. It may be a good idea to just have a test to try importing every function in the package, to avoid this issue in future.

Note: this issue comes to you as part of my wider [JOSS review](https://github.com/openjournals/joss-reviews/issues/2171)

RHammond2:
Closes #115 
Closes #116 

I've added readme and requirements file to examples folder so that people are notified that there are additional dependencies required for running the examples. I have also included an additional line in the main package readme.

Let me know if this will do the  trick for closing these two issues and if we should just update the readme in here for calling out python 3.6-3.8 only for using OpenOA (re: #114)


brynpickering:
Having thought about this some more, different requirements.txt files might not be the best way to solve this issue. Instead, perhaps a 'neater' alternative would be to use pip optional dependencies (see this [example](https://github.com/psf/requests/blob/3c1d36b827417fdeaf5a1c106129de30dac371d7/setup.py#L98)). That way, you could simply instruct users to install the package as per their intended use, e.g.:

```
pip install ./OpenOA[interactive]
```
Where the following lines have been added to `setup.py`:
```
extras_require={
          'interactive': ['matplotlib', 'jupyter', 'h5pyd']
      }
```

The same optional dependency groups could be used for docs and development installs, allowing you to condense all your requirements files into one place (`setup.py`).

Anyway, just a suggestion, you can also stick with the different `requirements.txt` files. Both work by me :-)

RHammond2:
I think the latter would be my preferred option (normally what I'd choose) because it's a bit cleaner and seems more intuitive, but isn't necessarily the approach that's been taken in this project to date. I'll defer to @jordanperr  on this one, and am happy to be the `setup.py` approach if he agrees or leave as-is otherwise.

RHammond2:
It looks like we're going to use the `requirements.txt` setup for now, but I will have a longer term issue opened up later that aims to have a more "elegant" solution to dependency management.

jordanperr (approved):

PR #127, opened 2021-02-01
Incorporate IAV in AEP calculation

nbodini:
This PR re-incorporates IAV in the AEP class (issue #86 ).

@ejsimley, feel free to check this out and tag Mike if you'd like to get his point of view on this, too.


operational_analysis/methods/plant_analysis.py
```diff
@@ -831,6 +832,11 @@ def run_AEP_monte_carlo(self):
                 inputs = inputs.reshape(-1,1)
             gross_lt = fitted_model.predict(inputs)
 
+            # Annual values of lt gross energy, needed for IAV
+            reg_inputs_lt['gross_lt'] = gross_lt
+            gross_lt_annual = reg_inputs_lt['gross_lt'].resample('12MS').sum().values
```
ejsimley, line 837: minor detail, but if the data has monthly time resolution I would think the normalization to 30-day months should be undone before calculating `gross_lt_annual`

ejsimley:
Hey @nbodini, overall this looks like a good approach to me. However, I still want to understand fundamentally what each element of `aep_GWh` is supposed to represent, so looping in @moptis here too. If an estimate of `aep_GWh` is intended to represent the expected AEP over a 20 year period (i.e., 20 year energy production / 20 years), then I don't see why IAV should be applied. It seems like it would be undoing the 20 year long-term average. On the other hand, if `aep_GWh` is supposed to represent the AEP for a single year, then I agree with how IAV is used. Does this make sense? 

In other words, is the uncertainty in `aep_GWh` that we report supposed to be the uncertainty in the expected AEP over 20 years, or is it supposed to be the uncertainty in the AEP we might observe in any single year?

moptis:
@ejsimley It'd be the latter, the AEP we'd expect in any given year, so IAV in this case is important.

operational_analysis/methods/plant_analysis.py
```diff
@@ -850,11 +856,20 @@ def run_AEP_monte_carlo(self):
...
+        # Calculate mean IAV for gross energy
+        iav_avg = iav.mean()
+        
+        # Apply IAV to AEP from single MC iterations
+        iav_nsim = np.random.normal(1, iav_avg, self.num_sim)
+        aep_GWh = aep_GWh * iav_nsim
+        lt_por_ratio = lt_por_ratio * iav_nsim
             
         # Return final output
         sim_results = pd.DataFrame(index=np.arange(num_sim), data={'aep_GWh': aep_GWh,
```
ejsimley, line 875: It might make sense to save the `iav` array in the sim_results. What do you think?

operational_analysis/methods/plant_analysis.py
```diff
@@ -831,6 +832,11 @@ def run_AEP_monte_carlo(self):
                 inputs = inputs.reshape(-1,1)
             gross_lt = fitted_model.predict(inputs)
 
+            # Annual values of lt gross energy, needed for IAV
+            reg_inputs_lt['gross_lt'] = gross_lt
+            gross_lt_annual = reg_inputs_lt['gross_lt'].resample('12MS').sum().values
+            #gross_lt_annual = gross_lt.reshape(-1, int(self._calendar_samples)).sum(1)
```
ejsimley, line 838: clean up comments

ejsimley (changes requested):
Looked at this in a little more detail and overall it looks good to me! Please see the comments I left, especially regarding undoing the normalization to 30-day months. 

Two more requests. 
1) Please update the documentation in example 02 where it says that IAV is not included yet.
2) Add a bullet point to the changelog in the "Unreleased" section (which needs to be added) briefly describing the addition of IAV. This will make it easier to keep track of changes when updating the version number later.

operational_analysis/methods/plant_analysis.py
nbodini: Yes, I like that idea. It's now incorporated.

operational_analysis/methods/plant_analysis.py
nbodini: Nice catch. Yes, I agree with you.

nbodini:
@ejsimley thank for your comments, I have now incorporated them in the PR.

ejsimley (approved):
This looks good now. I tested the daily linear regression and the uncertainty is now similar to monthly linear. Thanks for updating the examples!

PR #134, opened 2021-03-02
Fix power production outside valid operation windspeeds

### LINKED ISSUE #133: IEC Power Curve Not Behaving As Expected
https://github.com/NREL/OpenOA/blob/6bed784a50cb223386fce38afb1929365faa8f90/operational_analysis/toolkits/power_curve/functions.py#L13-L51

I was using the IEC power curve function, but found that the `windspeed_end` input does not always translate to a cutoff speed at which all power output is zero above a certain windspeed. There are a variety of different ways that it goes wrong, but I have found that in order for it to work correctly, the provided `windspeed_column` and `power_column` must have a zero power input as the last entry that corresponds to windspeed that is two `bin_width`s after the last windspeed.

This means that if `bin_width=0.5` and the last power-producing windspeed is 25m/s, then the windspeed entry corresponding to 0kWh must be 26m/s. If the the last windspeed and power combination is defined in any other way, then the power curve function returns a non-zero value for all windspeeds greater than 25m/s.

RHammond2:
Closes #133 
This patches a bug in the IEC power curve function to automatically set the power production to 0 for any windspeed that is outside of the (`windspeed_start`, `windspeed_end`) bounds.

It seems like we probably need to develop a better set of tests for this method, so let me know if you'd like to discuss this further, and so I haven't updated them at this point.


operational_analysis/toolkits/power_curve/functions.py
```diff
@@ -46,6 +47,8 @@ def pc_iec(x):
         for i in range(0, len(bins) - 1):
             idx = np.where((x >= bins[i]) & (x < bins[i + 1]))
             P[idx] = P_bin[i]
+        cutoff_idx = ((x < windspeed_start) | x > windspeed_end)
```
ejsimley, line 50: Missing parentheses around `x > windspeed_end`. Also, minor point, but maybe it should be `x >= windspeed_end` since all other bins are defined as open on their upper bound? But I think it's fine the way it is for now.

ejsimley (changes requested):
Overall this change looks good pending a minor edit. Definitely an improvement over the old method.

RHammond2:
@jordanperr I've just added a new IEC test, let me know if I should go further with this.

test/unit/test_power_curve_toolkit.py
```diff
@@ -16,13 +16,35 @@ def setUp(self):
         self.x = pd.Series(np.random.random(100) * 30)
         self.y = pd.Series(logistic5param(self.x, *params) + np.random.random(100) * noise)
 
+        # power curve source: https://github.com/NREL/turbine-models/blob/master/Offshore/2020ATB_NREL_Reference_15MW_240.csv
+        self.nrel_15mw_wind = pd.Series(np.arange(4, 26))
+        self.nrel_15mw_power = pd.Series(np.array([720, 1239, 2271, 3817, 5876, 8450, 11536, 1500, 1500, 1500, 1500, 1500, 1500, 1500, 1500, 1500, 1500, 1500, 1500, 1500, 1500, 1500]))
```
ejsimley, line 21: Looks like an extra 0 is missing from the rated power (should be 15000)

RHammond2: Good catch, that's an embarrassing copy paste error.

ejsimley (approved):
This looks ready to me.

jordanperr (approved):
- Merged develop
- Updated changelog
- Reviewed code changes
- Verified that regression tests pass

Issue #142, opened 2021-04-14
Shapely geos_c.dll not found


charlie9578:
When trying to run OpenOA/examples/02_plant_aep_analysis.ipynb I get the following error: 

_FileNotFoundError: Could not find module 'c:\users\charlie\anaconda3\envs\openoa-env\Library\bin\geos_c.dll' (or one of its dependencies). Try using the full path with constructor syntax._

This appears to be due to Shapely and the mix of using pip and conda: https://github.com/Toblerity/Shapely/issues/1032. The solution is to run: conda install shapely

Perhaps this will be resolved though in the next release of Shapely: https://github.com/Toblerity/Shapely/pull/1108

giswqs:
I am wondering when shapely v1.8 will be released. Shapely is a dependency of ipyleaflet. Currently, the ipyleaflet pip installation fails on Windows due to this `geos_c.dll not found` error. Would love to see this fixed. 

@davidbrochart @jorisvandenbossche @sgillies 

https://github.com/jupyter-widgets/ipyleaflet/issues/769
https://github.com/jupyter-widgets/ipyleaflet/pull/770

jordanperr:
@charlie9578 - Thanks for pointing out this issue. We haven't run into it ourselves, but I propose to add a small note in the README with special instructions for windows users to install Shapely using conda instead of pip. Does this sound like a workable solution? Would you kindly submit a pull request into the develop branch with the change to the readme? Thank you.

charlie9578:
Yes, that sounds good to me. Fingers crossed I've achieved a pull request!?

Issue #141, opened 2021-04-14
Add pywin32==225 to requirements (at least when using Python 3.8)?


charlie9578:
I get the following error when trying to run an example notebook having followed the install guidelines. I suspect this is a general problem with python 3.8, https://github.com/mhammond/pywin32/issues/1431, but may be worth adding pywin32==225 to the requirements file so others don't hit this error...

 import win32api
ImportError: DLL load failed while importing win32api: The specified module could not be found.

The solution is to use: pip install --upgrade pywin32==225

jordanperr:
Hi @charlie9578, thank you so much for reporting this issue. We are not experiencing this issue with our [Windows CI pipeline](https://github.com/NREL/OpenOA/runs/2196267929?check_suite_focus=true). I have read the thread on the pywin32 issue tracker, and it looks like this may be a specific issue with Python from the Microsoft Store instead of from the python.org installers. Could you please provide some more information about your development environment so I can try and re-create the problem. For example, Python version (and vendor), Windows version, Conda version, Pip version, and the steps you've taken to install OpenOA.

charlie9578:
Hmmm, interesting... I tried to repeat the error in a new environment and it doesn't occur. So perhaps a freak event, or one I'd need to re-create with a full reinstall of Anaconda.

Either way: Python 3.8.8 (default, Apr 13 2021, 15:08:03) [MSC v.1916 64 bit (AMD64)] :: Anaconda, Inc. on win32; MS Windows 10 Home, 10.0.19041 Build 19041; conda 4.9.2; pip 21.0.1. 

I used "conda create --name openoa-env python=3.8," cloned the OpenOA repo using GitHub desktop, navigated to use "pip install ./OpenOA," installed jupyter notebook "conda install -c conda-forge jupyterlab," ran "jupyter notebook," and then upon running the example notebook the error occured.

EricBLivingston:
I'm having the same issue with Python 3.9.4. Switching to v225 solved it for me, as well. I installed everything starting with miniconda from their site (no Windows store or anything like that)

jordanperr:
@charlie9578 has added a line in the Readme installation instructions about this issue and with the recommended fix to switch to v225.

PR #143, opened 2021-04-23
Update readme.md for Windows install

charlie9578:
Note for helping install in Windows if there is a problem with Shapely and the geos_c.dll. See issue #142 for more details.


charlie9578:
I've added a sentence to the readme to describe how to solve the win32 issue #141 too

Issue #95, opened 2020-08-25
Improving how long term losses are calculated


ejsimley:
In `MonteCarloAEP.calculate_long_term_losses`, long term availability and curtailment losses are calculated by finding the mean percent losses by month (or day), correcting for the number of days in each month, then finding the annual average of the loss categories. The average annual losses are then applied to the long-term gross energy. 

But by applying equal weight to the percent losses in each month/day this approach seems to miss the possibility of high percentage availability losses occurring during months with lower energy production. If this were the case, the total percentage losses over the year might be significantly lower than the average of the monthly percentages. 

Could this be changed so that long term losses are calculated using the average of the monthly losses in terms of absolute energy instead of percentages?

@moptis, @nbodini curious what you think about this.

moptis:
@ejsimley This is a good discussion and an important point. I think both approaches have limitations. I think the absolute energy approach is more risky to anomalously high or low energy production months. Like suppose we had less than 2 years of data and the only data for March was a really high wind resource month but with the usual losses. An absolute energy approach would penalize that month of March with higher losses in the long-term average, while the percentage-based approach wouldn't.

Maybe weighting the monthly percentage losses by the mean modeled energy production over the long-term period would be a good middle ground?

ejsimley:
That's a good point @moptis Your suggestion of weighting the monthly percent losses by the long-term corrected monthly gross energy does seem like a more robust approach. I'll look into how we might modify the code to use this strategy. Thanks!

nbodini:
Thanks for pointing this out, @ejsimley .
I've been playing with the data a bit in the past couple days, and I just realized that using losses as percentages can cause some issues at the different time resolutions used in the AEP calculation (monthly vs daily). Basically, we wouldn't get the same long-term % losses when looking at the same project but at different resolutions. 
Let's say we want to quantify the average availability loss for October 2017 for a given project. For how the class is written now, when the AEP calculation is performed at monthly resolution, the average availability loss for that month would be calculated as the absolute availability loss for that month divided by energy. On the other hand, when the calculation is performed at daily resolution, the average availability loss for October 2017 would be calculated as the average of the 31 daily rations between absolute availability and energy. And in general, the average of a bunch of ratios is not necessarily equal to the ratio of the averages.
And I've noticed this can have an impact as large as a 1% difference in AEP between monthly at daily resolution..

PR #147, opened 2021-05-05
New long term losses calculation

ejsimley:
This modifies `calculate_long_term_losses` and `sample_long_term_losses` in `plant_analysis.py` to address issue #95. In the original loss calculations, the average monthly percentage losses are averaged to estimate long-term availability and curtailment losses. But this could miss the possibility of high losses occuring in months with low energy production (like summer), in which case the average long-term losses would not be as high. The new method weights the average monthly percentage losses by the average monthly long-term energy to account for possible relationships between percentage losses and energy production, following the suggestion by @moptis.

@nbodini this does not address the point you brought up about percentage losses being different depending on the time resolution of the AEP analysis. We could try to address that here (e.g., by always making losses based on monthly resolution), or save that for later.


operational_analysis/methods/plant_analysis.py
```diff
@@ -545,20 +545,16 @@ def trim_monthly_df(self):
...
         Returns:
-            (tuple):
-                :obj:`float`: long-term annual availability loss expressed as fraction
-                :obj:`float`: long-term annual curtailment loss expressed as fraction
+            (None)
         """
-        df = self._aggregate.df
-        
-        days_year_lt = 365.25 # Number of days per long-term year (accounting for leap year every 4 years)
+        df = self._aggregate.df.rename_axis(None)
```
nbodini, line 557: Just a minor comment: why do we need the .rename_axis(None) here? Maybe let's add a little comment to clarify that?

operational_analysis/methods/plant_analysis.py
```diff
@@ -919,20 +907,28 @@ def sample_long_term_reanalysis(self):
...
+    def sample_long_term_losses(self, gross_lt):
         """
         This function calculates long-term availability and curtailment losses based on the Monte Carlo sampled
-        historical availability and curtailment data
+        historical availability and curtailment data. To estimate long-term losses, average percentage monthly losses 
+        are weighted by monthly long-term gross energy.
         
         Args:
-            n(:obj:`integer`): The Monte Carlo iteration number
+            :obj:`pandas.Series`: Time series of long-term gross energy
```
nbodini, line 918: Again a minor thing, I've noticed that in some other functions in the code we use something like the following: gross_lt(:obj:`dataframe`) instead of :obj:`pandas.Series`. Is there a preferred convention and do we want to stick with that one?

nbodini (changes requested):
This looks good to me overall! Just a couple minor inline comments - see below. 
Also, I have another comment related to the example notebook. Right now, Step 7 calculates and prints:
print(avail.mean())
print(curt.mean())
This looks a bit confusing to me, as we are just showing the arithmetic average of the losses, while then we actually perform a weighted average of the losses in the AEP calculation. I guess someone could get confused on why in Step 7 the mean curtailment is about 0.09%, but then in the plot in Step 10 it becomes 0.06%. Maybe let's either add a sentence or two to clarify this, or we should change a bit how we show results in Step 7.

ejsimley:
@nbodini thanks for taking a look at this so quickly. As far as your comment on step 7 of the example notebook, I also wasn't sure how to handle this and I'm not really happy with the current format. Since the LT losses require the full AEP analysis before they can be calculated now, I don't know if it makes sense to still include this step. One option could be to plot the monthly average losses (similar to the plot in step 5, but for 12 monthly averages) and explain in the text that they will be weighted by gross energy in a later step to determine average LT losses. Do you think this makes sense?

nbodini:
@ejsimley yes, I agree. Either option works for me (either excluding Step 7 or showing the plot you suggested), as there would be no inconsistence with the distributions shown later in the example. Feel free to pick whatever you prefer :)

operational_analysis/methods/plant_analysis.py
ejsimley: this was needed to combine with the gross_lt Pandas Series in sample_long_term_losses. Multiplying them wouldn't work unless they had the same axis names (and gross_lt didn't have a name). But I decided to change it to keep the axis name 'time' and instead renamed the gross_lt axis to 'time' in sample_long_term_losses (with a comment there too)

operational_analysis/methods/plant_analysis.py
ejsimley: Yes, that does seem to be the most commonly used style for args. I changed it.

ejsimley:
Hey @nbodini, I addressed your comments and decided to remove the LT losses step from the example notebook. Let me know if that all looks good.

nbodini (approved):
Everything looks good to me! Thanks for addressing my comments, @ejsimley

RHammond2 (approved):
Code-wise, I think this looks good, and I definitely appreciate the addition of more in-line comments!

PR #148, opened 2021-05-14
AEP: add option for hourly resolution

nbodini:
This PR adds an option to perform the AEP calculation at hourly resolution. The 2b example is also changed to present the uncertainty reduction when the AEP is performed at hourly resolution.


nbodini:
Hmmm, the unit tests keep being cancelled, not sure why.. any ideas?

operational_analysis/methods/plant_analysis.py
```diff
@@ -89,12 +89,12 @@ def __init__(self, plant, reanal_products=["merra2", "ncep2", "erai","era5"], un
...
-            raise ValueError("time_res has to either be M (monthly, default) or D (daily)")
+        if time_resolution not in ['M','D','H']:
+            raise ValueError("time_res has to either be M (monthly, default) or D (daily) or H (hourly)")
         self.time_resolution = time_resolution
-        self._resample_freq = {"M": 'MS', "D": 'D'}[self.time_resolution]
-        self._hours_in_res = {"M": 30*24, "D": 1*24}[self.time_resolution]
-        self._calendar_samples = {"M": 12, "D": 365.25}[self.time_resolution]
+        self._resample_freq = {"M": 'MS', "D": 'D', 'H': 'H'}[self.time_resolution]
+        self._hours_in_res = {"M": 30*24, "D": 1*24, "H": 1}[self.time_resolution]
+        self._calendar_samples = {"M": 12, "D": 365, "H": 365*24}[self.time_resolution]
```
jordanperr, line 97: Why was daily changed from 365.25 to 365?

jordanperr:
This pull request looks simple enough to me. Fine merging it after we discuss the change in the length of a year from 365.25 to 365. Why was this change made?

Looks like Github Actions have been experiencing degraded performance. You can check the status here: https://www.githubstatus.com We can ignore these failing actions (for now) by running the tests locally and confirming they pass.

operational_analysis/methods/plant_analysis.py
ejsimley: Adding to Jordan's comment, it seems like this changes (slightly) how many days will be used when long-term sampling the reanalysis data with daily resolution (ignoring leap years). Is there a reason for this?

ejsimley (changes requested):
This looks good overall. Happy to see the changes are relatively minor to implement hourly resolution! I do have some comments on example 02b. 
- At the beginning of Section 1.1, I think it would be worth mentioning that the GAM model will use hourly resolution, while GBM and ETR will use daily. 
- The documentation at the beginning of Section 1.2 still says that the multivariate GAM model has daily resolution, but it was changed to hourly in the example.
- The last cell explains that uncertainty is reduced when using the multivariate GAM, but compared to the univariate GAM model, uncertainty actually increases. Do you want to change how you're explaining the results? Or are more iterations required to converge to lower uncertainty?

operational_analysis/methods/plant_analysis.py
nbodini: I had changed this because of two reasons:
1) the QC on availability and curtailment (lines ~572) was not working well at hourly resolution with the 365.25 choice. Basically, that QC step would always fail for not-leap years.
2) using 365.25 was messing up some of the IAV calculations. Basically, since 365.25 is an approximate value, the length of the long-term reanalysis dataset was not exactly 100% accurate. At fine (mostly hourly) resolution we sometimes were ending up with a few extra hours, that were being treated as a full new year during the IAV calculation, and this was leading to an extra year with very low production (just a few hours of data), which led to very large IAV.
Setting self. _calendar_samples  to 365 fixes 1), and it worked well for 2) too, even though it is not a perfect solution, either, as we're losing a few timestamps in the creation of the long-term reanalysis dataset, which would now be N years minus a few hours.

I've thought more about this, and I think the following solution works better to build the long-term reanalysis dataset, as it exactly selects the last N years of data as wanted:
        long_term_reg_inputs = ws_df[ws_df.index[-1] + ws_df.index.freq - pd.offsets.DateOffset(years = self._run.num_years_windiness):]
With this new way to create the LT reanalysis dataset, we can then stick with the self. _calendar_samples at 365, as that will now only be used for the availability and curtailment QC step.

The one issue I have now is that the IAV still is not working as expected with the long-term solution proposed above. The issue is connected to this command:
gross_lt_annual = reg_inputs_lt['gross_lt'].resample('12MS').sum().values
Let's say that we have an 11-year period (at hourly resolution) created with the more accurate offset command above that starts at 2008-12-31 01:00:00, and ends at 2019-12-31 00:00:00. Now, this is exactly 11 yers, so the newly-proposed command above works well. However, the resample command above will actually create 12 years of gross_lt_annual, instead of the expected 11 years. This is because the resample command starts from December 1 in creating the annual values (so each year will go Dec 1 - Nov 30). So year 1 would lose data for 1-30 December 2008, and we'd get an additional 12th year with just data from 1-30 December 2019. 
Any suggestions on how to fix this issue? I have looked at other ways to use the resample command here but could not find an easy fix.

nbodini:
Thank you @jordanperr and @ejsimley for your comments.
 
As for the comments on the example, I have fixed the first two comments. For the last comment, I was actually seeing a decrease in uncertainty when going from univariate hourly gam to multivariate hourly gam (7% -> 6.6%). I guess that might not be always the case given the relatively small number of MC iterations. 

After we agree on how to fix the issue connected to the LT correction, I will increase the number of MC iterations to 500 to try and have more consistent results between different runs.

operational_analysis/methods/plant_analysis.py
nbodini: Okay, I think I have now fixed the remaining issue, see lines 844-847. It's not very elegant, but it does its job. Happy to receive suggestions on how to make this chunk of code nicer.

operational_analysis/methods/plant_analysis.py
```diff
@@ -841,7 +841,10 @@ def run_AEP_monte_carlo(self):
             
             # Annual values of lt gross energy, needed for IAV
             reg_inputs_lt['gross_lt'] = gross_lt
-            gross_lt_annual = reg_inputs_lt['gross_lt'].resample('12MS').sum().values            
+            gross_lt_annual = np.zeros(self._run.num_years_windiness)
+            # Annual resample starting on the first day in reg_inputs_lt
+            for j in np.arange(self._run.num_years_windiness):
+                gross_lt_annual[j] = reg_inputs_lt['gross_lt'].loc[(reg_inputs_lt.index >= reg_inputs_lt.index[0] + j*pd.tseries.offsets.DateOffset(months = 12)) & (reg_inputs_lt.index < reg_inputs_lt.index[0] + (j+1)*pd.tseries.offsets.DateOffset(months = 12))].sum()
```
ejsimley, line 847: This looks like it is working as expected, although I wonder if there is a more elegant way to compute this without a for loop. One thought is could you still use the resample('12MS') method but first shift the time index so that it starts at the beginning of the month? Maybe @jordanperr has some ideas here.

ejsimley (changes requested):
@nbodini It looks to me like you have solved the issues related to sampling exactly an integer number of years and calculating IAV correctly. Good catch finding the problem with the partial years contributing to erroneously high IAV!

I agree with you that your solution to IAV is correct but no longer as elegant. I couldn't quickly come up with a better method, though. As I commented in the code, maybe Jordan has a good idea how to improve the code. Perhaps there is a way to shift the time index so that it starts on the beginning of the month and then we could use `resample('12MS')`. But I wouldn't let this hold up the pull request.

A couple minor issues with the example notebook still.
-  The GAM model in Section 1.1 uses temperature whereas it should be a univariate regression
- Even after I changed the GAM model in section 1.1 to univariate, I still saw the same uncertainty when temperature is added (7.7% with 2000 iterations). GAM with ws and temperature has higher P50 AEP though.
- Very minor grammar correction in the last cell

nbodini:
Thanks, @ejsimley for catching the additional issues in the example. I have reviewed the example, and corrected the first gam setup and the typo in the last section.
Now I see a 7.7% uncertainty for the univariate GAM, and a 7.6% when adding temperature too. I think this very reduced change in uncertainty might be consistent with the results we found in the Wind Energy paper, i.e., that adding temperature has large benefits for wind farms with a strong seasonal cycle. Maybe this one simply did not experience one? If that's the case, I have added a comment in the last section of the example to suggest this and refer the reader to our Wind Energy paper to see additional results.

operational_analysis/methods/plant_analysis.py
ejsimley: Here is another method you could consider that seems to work correctly. It avoids a for loop, but still involves a lot of code:
```
reg_inputs_lt_temp = reg_inputs_lt['gross_lt']
# now shift time index to beginning of first month so resampling by 'MS' works as expected
reg_inputs_lt_temp.index = reg_inputs_lt_temp.index - pd.Timedelta(reg_inputs_lt_temp.index[0] - (reg_inputs_lt_temp.index[0].floor('d') + pd.offsets.MonthEnd(0) - pd.offsets.MonthBegin(1)))
gross_lt_annual = reg_inputs_lt_temp.resample('12MS').count().values
```

ejsimley: Either way, these long lines should be split into multiple lines to adhere to the 120 character linewdith

ejsimley (approved):
The example notebook changes look good to me Nicola! I approve this, pending any coding style suggestions @jordanperr might have, especially around line 846 in `plant_analysis`

operational_analysis/methods/plant_analysis.py
RHammond2: I think that this functionality could be its own function in the `toolkits/timeseries.py` module with caching because we're going to recompute the same thing a number of times as we simulate this, but if we don't want to go that route, the solution you have Eric is what I'd recommend. I tried to get really slick about it, but pandas doesn't want to play nice in a way that makes things easy and clean. That said, I'd break it up into lines like the following (adjust for proper indents after copy/paste):
```python
reg_inputs_lt_temp = reg_inputs_lt['gross_lt']
# now shift time index to beginning of first month so resampling by 'MS' works as expected
reg_inputs_lt_temp.index = (
    reg_inputs_lt_temp.index = (
    reg_inputs_lt_temp.index - pd.Timedelta(
        reg_inputs_lt_temp.index[0]
        - (
            reg_inputs_lt_temp.index[0].floor('d')
            + pd.offsets.MonthEnd(0)
            - pd.offsets.MonthBegin(1))
        )
)
)
gross_lt_annual = reg_inputs_lt_temp.resample('12MS').sum().values
```

ejsimley: @RHammond2 and @nbodini I made a separate function `get_annual_values` to calculate the annual values of LT gross energy and use this in the AEP analysis now. Please take a look and see what you think.

operational_analysis/methods/plant_analysis.py
```diff
@@ -28,6 +28,36 @@
...
 
+
+def get_annual_values(data):
+    """
+    This function returns annual summations of values in a pandas Series (or each column of a pandas DataFrame) with a
+    DatetimeIndex index starting from the first row. The purpose of the function is to correctly resample to annual
+    values when the first index does not fall on the beginning of the month.
+
+    Args:
+        data(:obj:`pandas.Series or pandas.DataFrame`): Input data with a DatetimeIndex index.
```
ejsimley, line 39: @RHammond2 Do you know how to indicate that multiple inputs are allowed? Not sure if this is right.

operational_analysis/methods/plant_analysis.py
nbodini: Looks good to me! And thanks for doing this.

operational_analysis/methods/plant_analysis.py
```diff
@@ -28,6 +28,36 @@
...
+    data.index = (
+        data.index - pd.Timedelta(
+            data.index[0]
+            - (
+                data.index[0].floor('d')
+                + pd.offsets.MonthEnd(0)
+                - pd.offsets.MonthBegin(1)
+            )
+        )
+    )
```
RHammond2, line 56: Because it's in a function I might just rewrite it as the following just for the sake of clarity in what is going on in the code:
```python
    ix_start = data.index[0]
    month_start = ix_start.floor("d") + pd.offsets.MonthEnd(0) - pd.offsets.MonthBegin(1)
    data.index = data.index - pd.Timedelta(ix_start - month_start)
```

RHammond2 (approved):
Updating the `get_annual_values()` function is up to you but I'd at least make the one change to the docstring. Otherwise, this is good to go.

operational_analysis/methods/plant_analysis.py
ejsimley: Thanks Rob, I made both of those changes you suggested.

Issue #139, opened 2021-04-12
Update Sphinx config py to read version from __init__.py


jordanperr:
- [x]  The version is hard coded into conf.py https://github.com/NREL/OpenOA/blob/main/sphinx/conf.py#L82
Version is defined here: https://github.com/NREL/OpenOA/blob/6bed784a50cb223386fce38afb1929365faa8f90/operational_analysis/__init__.py#L1

jordanperr:
Closed by #152, thanks @charlie9578 !

PR #152, opened 2021-05-19
#139 update sphinx config py to read version from   init  .py

charlie9578:
Importing the version number from the operational_analysis init file and using it in the sphinx conf file to fix issue #139


jordanperr:
Thanks, Charlie, for this pull request. The changes look good to me.

Issue #149, opened 2021-05-18
Create a wind farm map based on the project-asset information


charlie9578:
I'd like to see some additional plotting tools for analysts to use, so we can better understand the assets. As a starter I'm proposing to add a wind farm map using the project-asset data. A couple of things I'd like to check:

1. Is it okay to add the function to the pandas_plotting.py toolkit?
2. Is it okay to use the [bokeh library](https://docs.bokeh.org/en/latest/index.html)? I can add it to the requirements
3. Is it okay to add an example to the first [Example notebook](https://openoa.readthedocs.io/en/latest/examplesout.html#Import-the-data), to highlight the asset data, above the turbine list output?

ejsimley:
Charlie, glad that we could merge in your wind farm map plotting contribution, thanks so much for your work on that! Is it OK to close this issue? Also, happy to discuss other feature ideas you're interested in sometime.

charlie9578:
Hi Eric, glad to see the maps now included! So yes, safe to close. I'll have a think about what to work on next. Thanks, Charlie

PR #153, opened 2021-05-19
Feature/149 windfarm map

charlie9578:
Adding a wind farm plot as described in issue #149. I've also updated the sphinx documentation and requirements to facilitate bokeh plots, and included the plot as an example in the 00_toolkit_example notebook.


charlie9578:
I seem to sometimes fail the "docs/readthedocs.org:openoa" tests when creating a pull request, but they all pass locally. In this case I think its due to the packages, and potetially the docs test only installing packages in the sphinx/requirements.txt, rather than those in the main requirements. 

Is there a way to also run the test on my forked github to check it passes before pushing to the NREL repo?

jordanperr:
Hi Charlie, you can make an account with readthedocs.org and point it to your own fork of OpenOA. I'm happy to walk you through the configuration. Readthedocs installs only a subset of the dependencies which are required to build the documentation. Historically, a full installation of OpenOA was not needed for the documentation.

operational_analysis/toolkits/pandas_plotting.py
```diff
@@ -710,3 +714,204 @@ def turbine_polar_contour(
...
+    """Plot the windfarm spatially on a map using the Bokeh plotting libaray.
+
+    Args:
+        project(:obj:`plant object`): project to be plotted
+        tile_name(:obj:`str`): tile set to be used for the underlay, e.g. OpenMap, ESRI, OpenTopoMap
+        plot_width(:obj:`scalar`): width of plot
+        plot_height(:obj:`scalar`): height of plot
+        marker_size(:obj:`scalar`): size of markers
+        kwargs_for_figure(:obj:`dict`): additional figure options for advanced users, see Bokeh docs
+        kwargs_for_marker(:obj:`dict`): additional marker options for advanced users, see Bokeh docs. We have some custom behavior around the "fill_color" attribute. If "fill_color" is not defined, OpenOA will use an internally defined color pallete. If "fill_color" is the name of a column in the asset table, OpenOA will use the value of that column as the marker color. Otherwise, "fill_color" is passed through to Bokeh.
```
jordanperr, line 801: I added some extra documentation for the "fill_color"  parameter, for which some custom behavior is defined on lines 867-898. @charlie9578 , could you please verify my description is correct and update it as appropriate.

RHammond2 (changes requested):
Hi Charlie,

It looks like there is no need for the bokeh_plot.py module because you enabled the `bokeh_plot` extension functionality in conf.py, so it would probably be best to remove that file. Otherwise, this is a great addition, especially with the docstring plotting demonstrations.

Thanks!

jordanperr:
Hi Charlie, In addition to rob's suggestion, I added some more documentation about the "fill_color" option to the plotting function. I also loosened the bokeh version to 2.3.*, since fixing version numbers precisely can make the installation too brittle. I also reverted some changes in the tags-to-pypi and paper/* files. These must have been remnants of a hotfix that didn't make it all the way into your branch.

The tests are passing for me locally, but for some reason they are failing through gh actions. Upon inspection, it looks like gh-checkout is loading the wrong commit - cd30d01d66156a6fb446f06d31c932084a9ec59c instead of 2a7be1f8b885da8890a6d70d8d1585c1dc2d4777. I am not sure why this is happening. I might wait a day and try again in case it's a gh-actions bug.

operational_analysis/toolkits/pandas_plotting.py
charlie9578: Hi @jordanperr, thats the correct description. Thanks for cleaning up the code and checking things.

charlie9578:
> 
> 
> Hi Charlie,
> 
> It looks like there is no need for the bokeh_plot.py module because you enabled the `bokeh_plot` extension functionality in conf.py, so it would probably be best to remove that file. Otherwise, this is a great addition, especially with the docstring plotting demonstrations.
> 
> Thanks!

Hi @RHammond2, I've just tested removing the bokeh_plot.py module and like you say it seems to work fine. I've pushed the update and removed a couple of debugging print statements to clear things up as well. Hopefully all good now. Thanks, Charlie

RHammond2 (approved):

jordanperr (approved):

PR #164, opened 2021-06-19
Move gross energy description to proper section

abbey2017:
Minor change to the placement of the gross energy variable.


ejsimley:
Thanks for making this fix!

Issue #161, opened 2021-06-10
SVM algorithm missing from machine_learning_setup.py


heitkamz:
The docstring in machine_learning_setup.py notes that the module includes four algorithms to analyze: randomized tress, gradient boosting, generalized additive model, and support vector machine.

However, support vector machine is not an included algorithm in the body of the constructor.

nbodini:
Thank you for noticing this! We have now addressed this inconsistency with PR #165

PR #165, opened 2021-06-22
Remove SVM from description (issue #161)

nbodini:
Reference to support vector machine algorithms is now removed from the machine learning setup class.

The actual implementation of SVMs was removed by Mike in March 2019, and he says this was done because it was SUPER slow and not very useful, at least under that implementation - hyperparmeter optimization took forever.


ejsimley:
Thanks for fixing this Nicola.

PR #155, opened 2021-05-22
refactor code for better readability and using more pythonic convention

abbey2017:
#### Main Issues
- Code readability can be improved.
- PEP 8 style compliance can be enhanced according to project contribution guidelines.
- Raise overall coding standard of project.

#### Implementation
- Refactored boolean comparison and singleton check implementation in code.
- Removed trailing white spaces.
- Shorten length of some lines without reducing readability.

#### Remarks
All tests passed after code refactoring.

Issue #159, opened 2021-05-28
Electrical losses module can be refactored for better readibility


abbey2017:
#### Main isue
I would like to improve the readability of some OpenOA modules by refactoring the scripts according to PEP8 style starting with the electrical losses module.
Pylint score (max-line-length=120) of original script is -4.76/10.

The following issues were fixed with pull request #155:
- Proper boolean comparison and singleton check according to meet PEP8 style.
- Trailing white spaces
- Lines of code longer than 120 columns

#### Actions
- I used [autopep8](https://pypi.org/project/autopep8/) library (set to level 2 aggressiveness) to remove all trailing white spaces as follows: `autopep8 --in-place --aggressive --aggressive .\electrical_losses.py`
- Changed singleton check from `==` to  `is` according to PEP8 style
- Removed `==` from boolean comparision according to PEP8 style

#### Remarks
- The resulting script passed all tests.
- Pylint score of script increased from -4.76/10 to 5.85/10.

jordanperr:
Thanks for creating this detailed issue, @abbey2017, and for providing an example of electrical_losses.py linted in the manner you suggest. I have placed this issue in the V2.3 milestone and we will target a merge by then. As of PR #128 , linting with flake8 is now configured as a default pre-commit hook. We may combine your pull request #155  with some other work which we are doing internally to upgrade the code style in the codebase with flake8. Thanks again. No further action is required from you at this time.

PR #169, opened 2021-08-04
pre-commit all code

RHammond2:
This PR address the core concept in #155, but the pre-commit hook is nowapplied across the entire code base and nearly all linting/bug issues are addressed. Thank you so much @abbey2017 for highlighting that we missed running the pre-commit hooks over the entirety of the codebase when we implemented the checks, it's greatly appreciated!

All tests run, except one regression test: 
  `test/regression/int_pruf_plant_analysis.py::TestPandasPrufPlantAnalysis::test_daily_gbm`, 
but I'm curious if this has to do with regression testing a gbm model and we're looking at too fine-grained of results.

The few items that remain relate to:
 1) the executables hook in the following screenshot
image
 2) A typo (assuming) that gets caught by the linter, but I'd like confirmation on how this should be written. I assume that `df` should be `self.df`, but another question about the input `to` is being raised: should this be a valid input and is `to` really the target object to be copying from? Any help from @jordanperr or @ejsimley  on the intent here would be great.
    `operational_analysis/types/timeseries_table.py:340:38: F821 undefined name 'df'`
    the code itself:
    ```python
        def copy_column(self, to, fro):
            self.df = self.df.withColumn(df[fro])
    ```


abbey2017:
Hi Rob,
It seems like there is no action required from my side. Right?

Regards,
Abiodun

RHammond2:
Yes, that is correct, Abiodun! I just wanted to credit/recognize you and your prior PR in some way.

abbey2017:
Awesome, thanks!

Regards,
Abiodun

operational_analysis/methods/electrical_losses.py
```diff
@@ -1,261 +1,293 @@
...
             self._run = self._inputs.loc[n]
-            
+
             meter_df = self._plant._meter.df
-              
+
             # If monthly meter data, sum the corrected daily turbine energy to monthly and merge with meter
             if self._monthly_meter:
-                
-                scada_monthly = self._scada_daily.resample('MS')['corrected_energy'].sum().to_frame()
```
ejsimley, line 229: Out of curiosity, what doesn't it like about this line, aside from the quotes? It fits within the line width limit.

ejsimley (approved):
For the files I checked, this all looks fine to me. Thanks for taking care of this Rob. Once those points you mentioned get resolved, fine with me to merge.

operational_analysis/methods/electrical_losses.py
RHammond2: I just checked, and the line is 103 characters long with the indentation, which is beyond the 100 character limit, though I think the setting used to use 122 characters.

RHammond2:
@ejsimley I was just running the tests on the develop branch on my computer, and they also don't work there, though it seems that the GitHub Actions test running is working, so we can probably ignore what's happening on my end for the time being. Let me know if you have any last items on this.

operational_analysis/methods/electrical_losses.py
ejsimley: Hey Rob, the contributor guidelines file says that we use a line width of 120 characters: "This code follows the PEP 8 style guide and uses the pycodestyle linter to check for compliance. The only exception is the line length limit of 120 characters."
```
pylint --max-line-length=120 operational_analysis
```

Is there a reason you changed it to 100 characters?

ejsimley:
@RHammond2 The only last item I have is just to make sure we're consistent with the line width limit in the contributing documentation and the code. Otherwise, fine with me to merge.

operational_analysis/methods/electrical_losses.py
RHammond2: I think in the original PR, that had gotten merged before this step Jordan and I had gone back and forth a bit on line limits and decided on 100 as opposed to 122, but I think because that had been merged before the rest of the code got updated, that the documentation was not updated.

Would you prefer to go back to 122, or should I update the documentation to reflect this? Either way, thanks for catching this!

ejsimley: I'd say if you and Jordan decided on 100, then go with that. So, just go ahead and update the documentation in "contributing.md". Thanks!

RHammond2: Sounds great, thanks!

RHammond2:
@ejsimley I updated the Coding Style section to expand on our pre-commit workflow, would you like to see more details on our specifics in there, or do you think what I've amended is enough information?

ejsimley:
Hey Rob, the coding style section is pretty clear, thanks for updating that. Should we add that developers should activate the "pre-commit" workflow before committing their code just to be more explicit about that?

Also, looking at the [Readme file](https://github.com/NREL/OpenOA/tree/develop#Development), it comes across to me like if OpenOA is installed using `pip install -e ./OpenOA[develop]` then it is not necessary to activate the pre-commit workflow before committing using `pre-commit install`. Is that correct? Or is it always necessary to use `pre-commit install`?

RHammond2:
I've updated that section as well to be a bit more clear and consistent with the contributing guide. I'm going to go ahead and merge this now after the tests run.

ejsimley:
Looks good @RHammond2. Shall we close PR #155 and issue #159 then?

PR #189, opened 2022-06-26
improved compute_shear & added extrapolate_windspeed and test

zheitkamp1:
**Overview**
There are two material updates in this commit. The first is a revised version of compute_shear that is much more efficient and straightforward. The second is the addition of a function to extrapolate wind speed vertically. Both of these updates are within the met_data_processing.py module.

**compute_shear**
The current version of compute_shear is very slow. Using a Surface Laptop 4 (AMD Ryzen 7), a dataset containing 65,000 windspeed measurements (~7.5 years of hourly data) at 4 different measurement heights took 95 seconds to process! This is likely a result of the current function utilizing a for-loop where a new model for alpha/shear is fitted in each iteration.

Fortunately, since estimation of shear is a matter of simple OLS regression, shear can be computed directly with array operations. The proposed version of compute_shear is much faster (processing time of 0.02 seconds for the same dataset), more condensed, and arguably, more transparent. Another nice feature in the proposed compute_shear is efficient handling of NaN or 0 entry values.

**extrapolate_windspeed**
extrapolate_windspeed is a simple function that takes timeseries wind speed measurements with a corresponding set of timeseries shear and extrapolates the windspeed from the measurement height to a different height according to the Power Law. A unit test has been added for this function.


RHammond2:
Hi @zheitkamp1, thanks so much for submitting this pull request, and apologies for the delay in response, somehow none of us got any notifications regarding this PR. We've taken a quick look at this, and are quite excited for the refinement of our methods, and will start our formal reviews over the next week to provide any feedback that is necessary.

RHammond2 (approved):
Thanks again for this PR! I think this all looks good from my perspective, but will let @ejsimley ensure that the math is correct, though the tests seem to pass without modification of the expected rsults, which is a good sign.

ejsimley (changes requested):
Hi @zheitkamp1 thanks for contributing this pull request! The new implementation of the compute shear function is nicely written and very efficient. It's also nice to avoid having to specify a reference sensor when there are more than two heights. One comment, though, is that it isn't clear what reference height and wind speed a user should use if they wanted to extrapolate the wind speed with the resulting wind shear exponents. Could you return those reference values as well, or provide some guidance in the documentation? My understanding is that for a given sample with heights z and wind speeds u, the reference height and wind speed that correspond to the shear exponent are np.exp(np.mean(np.log(z))) and np.exp(np.mean(np.log(u))), respectively. Interested in hearing your thoughts!

zheitkamp1:
All - thanks for the feedback! @ejsimley, good suggestion. I think there are a few options for handling the reference height and wind speed. 

**Option 1:**
Modify `compute_shear` to return a tuple where the first element is the array of shear exponents, the second element is the reference height (float, since this will be the same for each time period), and the third element is the array of reference wind speed. Then user can use tuple unpacking to get these elements, e.g. : 
`shear, ref_height, ref_wind_speed = compute_shear(data, ws_heights)`

Pros:
- Easy modification

Cons:
- Fundamental change in the data structure that is returned. This updated version would return a tuple instead of an array/series. This may disrupt workflows that have been built on top of this function.
- Needs documentation instructing user on how to use reference height and wind speed.
- From a design standpoint, this violates single responsibility principle. Modified function would now be computing and returning the reference height and wind speeds.

**Option 2:**
Rename `compute_shear` to `compute_shear_parameters` and return a tuple where the first element is the array of the shear exponent (alpha) and the second element is the scaler/intercept term (beta) from the power law model:
$$U(z) = \beta z^\alpha$$

Then modify `extrapolate_windspeed` to take alpha, beta, and a target height as parameters and use the model to compute shear.

Pros:
- No need to keep track of the reference height and wind speed
- The outputs from `compute_shear_parameters` would flow nicely as inputs to `extrapolate_windspeed`.

Cons:
- Similar to option 1, this mod would be disruptive to workflows that have already been built around `compute_shear`
- This design implicitly forces use of the best-fit curve for the wind speed extrapolation. This isn't necessarily a problem unless you want to extrapolate wind speed based on the curve that passes through a specific measurement height.

**Option 3:**
Leave `compute_shear` and `extrapolate_windspeed` as they are and write a new function (call it `compute_shear_reference_values` ?) to calculate the reference height and wind speed. This function could take the same input parameters as `compute_shear` and return a tuple of (ref_height, ref_wind_speed).

Pros:
- Shouldn't disrupt workflows built around `compute_shear`.
- Easy modification.

Cons:
- Needs documentation for users on 1) why this function exists and 2) how to use the outputs
- Now user needs to call three functions (instead of two) if they have a dataset and want to extrapolate windspeed. For example. `compute_shear_reference_values` -> (ref_height, ref_wind_speed), then `compute_shear` -> alpha, finally `extraploate_windspeed(ref_height, ref_wind_speed, alpha, target_height)`.

Summary
I could be persuaded into going with any of these options or a different option I didn't think of. @RHammond2  and @ejsimley  do any of these options standout?

Thanks!

ejsimley:
Thanks @zheitkamp1, I appreciate the thorough explanation of the different options you came up with and pros and cons! Another option that I like, based on Option 1, is to add an additional boolean argument to `compute_shear` to specify whether the reference heights and wind speeds will be returned, like `return_reference_parameters`. This could be False by default, in which case only the shear is returned. If it is True then the function would return what you described in Option 1. Let us know what you think about this approach. As you pointed out in Option 2, really there is only one other parameter that is needed to fully describe the power law, $\beta$, but returning a reference height and wind speed means that `extrapolate_windspeed` wouldn't have to be changed.

zheitkamp1:
@ejsimley - I like that idea. With this implementation, there is no change to the behavior of `compute_shear` unless the user specifically asks for it - which is great.

I just pushed the commit with this implementation. Also added tests for the reference height and reference values. Please review and let me know if you have any other suggestions.

Thanks!

ejsimley:
Hi @zheitkamp1 thanks for adding the optional reference values to the function return, and for writing the test! It looks great! I just added a description of the change to the changelog file and made one other minor documentation change.

ejsimley (approved):

RHammond2:
@ejsimley, I'm just noticing that this is pulling into main, should we switch it to develop before merging?

ejsimley:
Good catch @RHammond2 yes I think it should be merged into develop first.

jordanperr:
Hi @zheitkamp1, thank you again for your contribution to OpenOA! We are ready to merge in this pull request. However, we notice that we do not know your company, email address, or even your name! While this is not technically required per the contributing guidelines (for now), we would like to be able to keep in touch with our contributors and provide attribution where appropriate. Would you feel comfortable with either or both of the following:

- Please add your name to the bottom of the README in the contributors section. There is no order to this list, so just place yourself at the end.
- Please reach out to me directly via email (jordan.perr-sauer@nrel.gov) to provide me with your contact information.

If you would prefer to do neither of these, that's fine too, and we can merge in your pull request anyway.

Thank You,
Jordan

zheitkamp1:
Hi @jordanperr - absolutely! I've updated the readme. I also sent you an email with some additional contact info and background. Hope we keep in touch!
-Zach

jordanperr:
Thanks @zheitkamp1, I received the email. Upon final review, I realized that these changes would break backwards compatibility with compute_shear function. In order to preserve backwards compatibility, we decided to move your changes into a new method (for now) called compute_shear_v3. I will merge these changes now into the develop branch, and then we will remove the old compute_shear in favor of this one in the upcoming OpenOA V3 release. Thank you for your patience with us!

jordanperr (approved):

PR #190, opened 2022-06-30
Remove old plant data files and update module hierarchy to V3 api

jordanperr:
- Move toolkits -> utils
- Move methods -> analysis
- Move types.plantdata -> plant
- Add package-level import of plantdata, for example, `from openoa import PlantData`
- Deprecate openoa.types (remove __init__.py, but keep files for now)
- Allow VSCode to auto-refactor, but did not test the auto-refactor yet

Notes:
- Should we keep utils, or rename to tools or promote the submodules to package-level modules? Argument for promotion to package-level modules is that each tool would represent a bounded context.
- Tests will not pass, we need to update the tests to use PlantDataV3
- Unit tests need to be written for PlantDataV3

### DIFF
--- a/openoa/types/asset.py
+++ b/openoa/types/asset.py
@@ -1,0 +1,205 @@
+import importlib
+import itertools
+
+import numpy as np
+import pandas as pd
+from pyproj import Transformer
+from shapely.geometry import Point
+
+
+class AssetData(object):
+    """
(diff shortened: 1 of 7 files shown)


RHammond2 (approved):
Looks good, thanks for putting this through!

PR #193, opened 2022-07-14
Modularize the QA module

RHammond2:
This PR accomplishes the following:
- deletes the old QA module
- refactors all the old QA functionality into the new module `openoa/utils/qa.py` and `openoa/utils/pandas_plotting.py`
- Adds a new example to introduce how we create `project_ENGIE.py` using the QA methods and demonstrates some basic `PlantData` concepts (@ejsimley this is particularly where I'd like your input to see if this is going in the right direction and to assess what's missing/unclear/etc.)
- Modifies `project_ENGIE.py` to be a bit more clear and deal with some missing timestamps in the ERA5 reanalysis data
- Adds all (most?) of the docstrings and a few bug fixes to `openoa/plant.py`

Additionally, I pulled in the changes from #192 so I could use the latest iteration of the code without worrying about duplicating any efforts.


openoa/plant.py
```diff
@@ -902,86 +1050,92 @@ def __attrs_post_init__(self):
...
+
+        if self.curtail is not None:
+            time_col = self.metadata.curtail.col_map["time"]
+            self.curtail[time_col] = pd.DatetimeIndex(self.curtail[time_col])
+            self.curtail = self.curtail.set_index([time_col], drop=False)
+            self.curtail.index.name = "time"
+
+        if self.asset is not None:
+            id_col = self.metadata.asset.col_map["id"]
+            self.asset = self.asset.set_index([id_col], drop=False)
```
jordanperr, line 1122: set_index with drop=false will preserve a duplicate id column

jordanperr:
I am reviewing this pull request instead of #192, since Rob requested that I review 192 and 193 together. At the same time I am closing 192.

Initial Comments: I was unable to get either of the v3_00 notebooks to fully run. 

**00_v3_demonstration**: Code cell 5, returns the following error:
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-15-9476e55f001b> in <module>
      1 engie.analysis_type = "all"
----> 2 engie.validate()

~/src/OpenOA/openoa/plant.py in validate(self, metadata)
   1184 
   1185         Returns:
-> 1186             (:obj: `dict[str, list[str]]`): A dictionary of each data type and any
   1187                 columns that  don't match the required dtype and can't be converted to
   1188                 it successfully.

~/src/OpenOA/openoa/plant.py in _validate_frequency(self, category)
   1151         """Validates that the column names in each of the data types matches the mapping
   1152         provided in the `metadata` object.
-> 1153 
   1154         Args:
   1155             category (str, optional): _description_. Defaults to "all".

~/src/OpenOA/openoa/plant.py in frequency_validator(actual_freq, desired_freq, exact)
    116         return actual_freq in desired_freq
    117 
--> 118     actual_freq = "".join(filter(str.isalpha, actual_freq))
    119     return actual_freq in desired_freq
    120 

TypeError: 'pandas._libs.tslibs.offsets.Hour' object is not iterable
```

I don't fully understand this error, but perhaps actual_freq is expected to be a string, but has now been updated to a pandas.offsets object?




**00_v3_toolkit_example:** Code cell 3 returns the following error:
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-3-57c6afb109c3> in <module>
----> 1 show(pandas_plotting.plot_windfarm(project.asset,tile_name="OpenMap",plot_width=600,plot_height=600))

~/Documents/GitHub/OpenOA/openoa/utils/pandas_plotting.py in plot_windfarm(asset_df, tile_name, plot_width, plot_height, marker_size, kwargs_for_figure, kwargs_for_marker)
    896 
    897     # Create the bokeh data source
--> 898     source = ColumnDataSource(asset_df)
    899 
    900     # Create a bokeh figure with tiles

/opt/miniconda3/envs/openoa/lib/python3.8/site-packages/bokeh/models/sources.py in __init__(self, *args, **kw)
    192         if not isinstance(raw_data, dict):
    193             if pd and isinstance(raw_data, pd.DataFrame):
--> 194                 raw_data = self._data_from_df(raw_data)
    195             elif pd and isinstance(raw_data, pd.core.groupby.GroupBy):
    196                 raw_data = self._data_from_groupby(raw_data)

/opt/miniconda3/envs/openoa/lib/python3.8/site-packages/bokeh/models/sources.py in _data_from_df(df)
    237         else:
    238             _df.index = pd.Index(_df.index.values, name=index_name)
--> 239         _df.reset_index(inplace=True)
    240 
    241         tmp_data = {c: v.values for c, v in _df.items()}

/opt/miniconda3/envs/openoa/lib/python3.8/site-packages/pandas/core/frame.py in reset_index(self, level, drop, inplace, col_level, col_fill)
   5011                 # to ndarray and maybe infer different dtype
   5012                 level_values = maybe_casted_values(lev, lab)
-> 5013                 new_obj.insert(0, name, level_values)
   5014 
   5015         new_obj.index = new_index

/opt/miniconda3/envs/openoa/lib/python3.8/site-packages/pandas/core/frame.py in insert(self, loc, column, value, allow_duplicates)
   3758         self._ensure_valid_index(value)
   3759         value = self._sanitize_column(column, value, broadcast=False)
-> 3760         self._mgr.insert(loc, column, value, allow_duplicates=allow_duplicates)
   3761 
   3762     def assign(self, **kwargs) -> DataFrame:

/opt/miniconda3/envs/openoa/lib/python3.8/site-packages/pandas/core/internals/managers.py in insert(self, loc, item, value, allow_duplicates)
   1189         if not allow_duplicates and item in self.items:
   1190             # Should this be a different kind of error??
-> 1191             raise ValueError(f"cannot insert {item}, already exists")
   1192 
   1193         if not isinstance(loc, int):

ValueError: cannot insert id, already exists
```

This seems to be caused by a duplicate id column in the asset dataframe. I think this is caused by a conflict between plant.py:1122 - where set_index is called with drop=False, and bokeh ColumnDataSource, which seems to do a reset_index - creating a duplicate id column. I would recommend removing drop=False from set_index in plant.py, but would like to hear your input first.

openoa/plant.py
RHammond2: This was more or less me being unsure if we should keep the column or not, so happy to defer to your judgement on if we should drop the column after setting the index because I could go either way with it.

RHammond2:
@jordanperr the `ColumnDataSource` issue I am aware of and will be fixed once we switch the internal naming convention to the IEC 61400-25-2 naming convention where there will be no conflict in the use of `id` as index, so I'd vote for letting the issue persist until the new naming convention is implemented. Alternatively, we could use `t_id` or `asset_id` in its place, which should resolve the issue.

As for the frequency issue, this should have been resolved from swapping `.freq` with `.freqstr`, which appears to be in the code working for me, now that I've cherry picked the commit from another branch that addressed the issue. Would you mind updating and double checking?

openoa/plant.py
RHammond2: After seeing your later comment, I've removed `drop=False`.

PR #196, opened 2022-07-28
Enhancement: Create tests for the plant module

RHammond2:
This starts building up a test suite to validate the all of the methods and clases in openoa/plant.py, and fixes any bugs that are found in the process.

PR #197, opened 2022-08-20
Add the v2 `AssetData` methods to v3 `PlantData`

RHammond2:
This PR slightly refactors the v2 `AssetData` methods to be incorporated within the v3 `PlantData` class in a mildly more efficient manner. i've also added the changes from #196 to be incorporated in here to not duplicate any docstring updates, so that should be merged first.

I've left this as a draft to get any thoughts about adding in tests at this stage. It doesn't seem like there were any from v2. I'm also open to any suggestions for any new additions/modifications here from from @ejsimley.


openoa/plant.py
```diff
@@ -1378,6 +1655,148 @@ def update_column_names(self, to_original: bool = False) -> None:
...
+
+        # Maintain v2 compatibility of np.inf for the diagonal
+        distance = distance + distance.values.T - np.diag(np.diag(distance.values))
+        np.fill_diagonal(distance.values, np.inf)
+        return distance
+
+    def calculate_nearest_neighbor(
+        self, turbine_ids: list | np.ndarray = None, tower_ids: list | np.ndarray = None
+    ) -> None:
+        """Finds nearest turbine neighbors for turbines and nearest tower neighbors for met towers.
```
ejsimley, line 1746: This made me think the method was going to only find the nearest turbine for turbine assets and nearest tower for tower assets. But it looks like it finds both the nearest turbine and tower for all assets. If that's correct, maybe update the description.

openoa/plant.py
```diff
@@ -1378,6 +1655,148 @@ def update_column_names(self, to_original: bool = False) -> None:
...
+
+        nearest_tower = distance[ix_tower].values.argsort(axis=1)
+        nearest_tower = pd.DataFrame(
+            distance.columns.values[nearest_tower], index=distance.index
+        ).loc[ix, 0]
+
+        self.asset.loc[ix, "nearest_turbine_id"] = nearest_turbine.values
+        self.asset.loc[ix, "nearest_tower_id"] = nearest_tower.values
+
+    def nearest_turbine(self, id: str) -> str:
```
ejsimley, line 1778: This method and the next one seem to require `calculate_nearest_neighbor` to be called first. Should there be check if the nearest turbine and tower columns exist, and if not call `calculate_nearest_neighbor`?

ejsimley:
Hi @RHammond2, the asset data methods look good to me, aside from some minor comments I made. The plant module is getting very long though, but it also might be hard to separate the asset methods into their own module, so I think it's fine for now. I did add a couple more methods to the `AssetData` class in my `wake_losses` branch based on v2 (https://github.com/ejsimley/OpenOA/blob/wake_losses/operational_analysis/types/asset.py). These are `direction_matrix` and `get_freestream_turbines`. Do you think we should add those to this pull request, or wait and include as part of a larger wake losses PR?

RHammond2:
@ejsimley, that's correct the cleanup is underway! I'd also say that incorporation of the wake losses work into v3 might be better as a separate effort because the code isn't yet a part of the v2 develop.

ejsimley (approved):
Hi Rob, this all looks good to me now, so fine with me to merge. Thanks!

PR #198, opened 2022-08-22
Utilities Improvements

RHammond2:
I've started this PR to address a few improvements to the utilities subpackage:
- Remove most of the helper functions from `openoa/plant.py` into openoa/utils/plant_data.py` to improve the readability of the plant module. The main concern was that it was getting a little bit unruly, so I wanted to shift anything that seemed like a utility function out of the main module and into the utils subpackage.
- Enable methods operate on both `pd.DataFrame`s (given some column names) and `pd.Series` to improve usability depending on a user's needs. This doesn't create thin wrappers, but starts to implement a separate method that operates with highly similar logic as the original implementation.
- General v3 cleanup of the utils subpackage, such as docstring improvement, inline comment improvement, code efficiency improvements (where applicable), etc.

@ejsimley and/or @jordanperr  It'd be great to get some feedback on the changes to plant and filters module at this stage to see if this is headed in the right direction for v3, or if there are other preferred approaches.


RHammond2:
@jordanperr this is ready to check now, and here are a couple concrete things I'd like feedback on:
- Does the slick pandas code make sense throughout?
- Does the dataframe/series usage make sense and is it consistent enough?
- Should the clustering and binning methods in `openoa/filters.py` should be left as-is or modified to be more similar to the others that operate on dataframes/series, and if so, should that be done here or add it in another PR just for the sake of getting this included?
- Should this [test](https://github.com/RHammond2/OpenOA/blob/enhancement/plant_utils/test/unit/test_imputing_toolkit.py#L263-L270) be deleted, or do you believe that it should actually work?
  - My finding was that in the original the target index is set as the returned data's index, so the alignment column should essentially disappear in the results, but in this test it's suggesting otherwise despite the other tests checking on the original index.

openoa/types/plant.py
```diff
@@ -10,7 +10,7 @@
 from openoa.types import timeseries_table
 
 from .asset import AssetData
-from .reanalysis import ReanalysisData
+from ..utils.reanalysis import ReanalysisData
```
jordanperr, line 13: Let's remove these relative imports.

jordanperr:
I think this is great. You're on the right track with the slick pandas code. Some high level comments:

- [ ] Let's discuss moving the functions in utils.plant_data back into plant, or renaming utils.plant_data to be something more meaningful.
- [ ] Would be nice to abstract the dataframe->series->dataframe pattern as a decorator that can be unit tested, instead of adding tests for dataframes and series on each function.

openoa/types/plant.py
RHammond2: This is resolved here: https://github.com/NREL/OpenOA/pull/198/commits/e92782a485a0475d68cc60e36c126662a41181a0

PR #200, opened 2022-08-30
Update AEP Method for OpenOA V3 API

jordanperr:
This pull request updates the long term corrected monte carlo aep analysis method (formerly known as plant_analysis.MonteCarloAEP) to use the V3 PlantData API. Specific changes:

**DONE:**
- Changed all column keys to V3 versions.
- Changed all references to TimeseriesTable classes to direct dataframe access
- Updated tests to reflect new naming convention

**TODO:**
- Debug numeric output errors: The tests are still not quite matching. Must debug before merge.
- Refactor: The class is still lengthy. Should be possible to perform a substantial refactor. May be done before merge.


examples/project_ENGIE.py
```diff
@@ -228,7 +228,7 @@ def prepare(path="data/la_haute_borne", return_value="plantdata"):
         # Build and return PlantData
         engie_plantdata = PlantData(
             analysis_type="MonteCarloAEP",  # Choosing a random type that doesn't fail validation
-            metadata="data/plant_meta.yml",
+            metadata=f"{path}/../plant_meta.yml",
```
RHammond2, line 231: I think this would just be `f"{path.parent}/plant_meta.yml"`

RHammond2, line 231: Not sure where my original comment went, but here it is again. This could just be the following to work consistently across OSes:
```python
metadata=path.parent / "plant_meta.yml",
```

openoa/analysis/aep.py
```diff
@@ -74,21 +78,21 @@ class MonteCarloAEP(object):
...
-        end_date_lt=None,
-        reg_model="lin",
-        ml_setup_kwargs={},
-        reg_temperature=False,
-        reg_winddirection=False,
+        plant: PlantData,
+        reanal_products: list[str] = ["merra2", "ncep2", "erai", "era5"],
+        uncertainty_meter: float = 0.005,
+        uncertainty_losses: float = 0.05,
+        uncertainty_windiness: tuple[int] = (10, 20),
```
RHammond2, line 85: Just for reference, tuples have to be annotated as either `tuple[int, int]` in this case, or for a dynamic number of values: `tuple[int, ...]`.

PR #201, opened 2022-08-30
Fix bug in `iter_validator` that failed with `None` values

RHammond2:
This is a small, isolated fix to address an issue for the `PlantData.analysis_type` validator doesn't correctly accept None as valid input. This fix converts an input of `None` to be `type(None)` in the actual validation so that a passed value of `None` will correctly validate to be a valid input.

@jordanperr this will solve the one issue we discussed earlier, and now allows the regression tests to fail somewhere else because of unrelated issues.


RHammond2:
This PR now simplifies the `analysis_type` validation routine to be self-contained rather than relying on two separate methods, in addition to changing `self._asset` to `self.asset` in the `PlantData.parse_geometry()`

PR #212, opened 2022-09-29
Bring all Analysis Modules up to v3 spec with speed/readability improvements

RHammond2:


openoa/analysis/electrical_losses.py
```diff
@@ -3,290 +3,353 @@
...
-        Args:
-            (None)
-
-        Returns:
-            (None)
+        This data frame is stored as self.inputs.
         """
         if self.UQ:
+            n_decimal = (len(str(el.split(".")[1])) for el in self.uncertainty_correction_threshold)
+            integer_multiplier = 10**n_decimal
```
jordanperr, line 160: I can't get this line to work. I receive the error "TypeError: unsupported operand type(s) for ** or pow(): 'int' and 'generator'" which seems to be caused by raising  integer 10 to a generator. The pow operator is only defined for numpy arrays.

jordanperr:
I can not get the second electrical loss test to pass.

I have merged this with my aep refactor branch (lots of conflicts, I think we duplicated some work on the AEP class) but the electrical loss test failure remains. I used your changes almost always. The only change I kept from my branch is a new location for the logger, which should reduce circular import issues.

Here is a branch with the merge: https://github.com/jordanperr/OpenOA/tree/merge_rhammond_212_with_jordanperr_v3_aep_refactor

I didn't want to commit this into your branch yet due to the test failure. One thing at a time.

- [ ] Review the electrical loss test failure
- [ ] Check out my merge branch and bring that in here if you're happy with it.

openoa/analysis/electrical_losses.py
RHammond2: That's my fault, it's supposed to be `max(len...)`.

RHammond2:
Thanks, @jordanperr, I'll take a look at this on Monday before I get any further with this soon to be behemoth of a PR.

RHammond2:
@jordanperr I just checked out your branch, and it all looks good, so just go ahead and do a PR into here. I've also fixed the bug in `ElectricalLosses`, which was just a typo.

RHammond2:
@jordanperr I'm taking this out of the draft mode now, though there will be some small updates, particularly to the plotting that need to be addressed after some of the upstream PRs are merged. The only real concern I have at this point is that the regression tests for the TIE method are failing because the numbers a bit different.

PR #217, opened 2022-11-07
Adopt IEC 61400-25 tag naming convention for data variables

RHammond2:
This PR primarily address the IEC 61400-25 tag naming throughout the codebase. However, there seem to be some fields that are unaddressed such as timestamps and turbine energy. @ejsimley I was going to update some of those missing in the metadata classes in `plant.py`, but was realizing that it's likely not as straightforward as I'd assume, so if there are any others you're aware of from the ENTR work, please feel free to fill them in or add them in the comments. Additionally, in the analysis classes there are fields like energy_corrected, energy_imputed, etc., and wasn't sure if we should replace "energy" with the appropriate tag naming convention. 

In addition to the tag naming updates, this PR also does the following:
- Tidies up the package requirements after having some issues with a fresh environment creation and rerunning the code
- Updates failing tests (the AEP the and TIE tests fail due to ongoing numerical consistency issues)
- Fixes up any lingering missed issues from the RC1 release
- Adds more comments to the metadata YAML file to further helping guide users through the new naming convention


sphinx/api/types.rst
```diff
@@ -89,18 +89,18 @@ data's column names in their SCADA data, plus the "freq" field. Users just have
...
+==================== ================= ================================== =============================
+ time                 time stamp        datetime64[ns]                     datetime64[ns]
+ id                   id                string                             None
+ WTUR_W               power             float                              kW
+ WMET_HorWdSpd        windspeed         float                              m/s
+ WMET_HorWdDir        winddirection     float                              degrees
+ status               status            string                             None
+ WROT_BlPthAngVal     pitch             float                              degrees
+ WMET_EnvTmp          temp              float                              Celsius
+==================== ================= ================================== =============================
```
ejsimley, line 103: Nice table

sphinx/api/types.rst
```diff
@@ -271,18 +270,18 @@ data. Air density at the 58th model level is calculated using temperature data e
...
- temperature          float                                     Kelvin
- density              float                                     kg/m^3
- surface_pressure     float                                     Pa
-==================== =======================================   ==================================
+==================== ================== ======================================= ==================================
+ Field Name           Descriptive Name   Data Type (ReanalysisMetaData.dtypes)   Units (ReanalysisMetaData.units)
+==================== ================== ======================================= ==================================
+ time                 time stamp         datetime64[ns]                          datetime64[ns]
+ WMETR_HorWdSpd       windspeed          float                                   m/s
+ WMETR_HorWdSpdU      windspeed-u        float                                   m/s
```
ejsimley, line 278: to be more descriptive, could say "eastward wind speed component"

sphinx/api/types.rst
```diff
@@ -271,18 +270,18 @@ data. Air density at the 58th model level is calculated using temperature data e
...
+ WMETR_HorWdSpdV      windspeed-v        float                                   m/s
```
ejsimley, line 279: to be more descriptive, could say "northward wind speed component"

examples/data/cleansed/metadata.yml
```diff
@@ -0,0 +1,61 @@
...
+  merra2:
+    time: time
+    WMETR_HorWdSpd: WMETR_HorWdSpd
+    WMETR_HorWdSpdU: WMETR_HorWdSpdU
+    WMETR_HorWdSpdV: WMETR_HorWdSpdV
+    WMETR_HorWdDir: WMETR_HorWdDir
+    WMETR_EnvTmp: WMETR_EnvTmp
+    WMETR_AirDen: WMETR_AirDen
+    WMETR_EnvPres: WMETR_EnvPres
+    frequency: H
```
ejsimley, line 61: what is this cleansed metadata file used for?

examples/data/cleansed/metadata.yml
```diff
@@ -0,0 +1,61 @@
+scada:
+  time: time
+  id: id
+  WTUR_W: WTUR_W
+  WMET_HorWdSpd: WMET_HorWdSpd
+  WMET_HorWdDir: WMET_HorWdDir
+  status: status
```
ejsimley, line 7: From the -25 tag list, it seems like WTUR.TurSt is the right tag name for turbine status

examples/data/cleansed/metadata.yml
```diff
@@ -0,0 +1,61 @@
...
+  WROT_BlPthAngVal: WROT_BlPthAngVal
+  WMET_EnvTmp: WMET_EnvTmp
+  energy: energy
```
ejsimley, line 10: In ENTR, we're calling turbine energy WTUR.SupWh

examples/data/cleansed/metadata.yml
```diff
@@ -0,0 +1,61 @@
...
+  frequency: 10T
+meter:
+  time: time
+  MMTR_SupWh: MMTR_SupWh
+  frequency: 10T
+tower:
+  time: time
+  id: id
+  frequency: 10T
+status:
```
ejsimley, line 20: I'm not sure if -25 has tag names for status codes. We could leave this as is for now.

examples/data/cleansed/metadata.yml
```diff
@@ -0,0 +1,61 @@
+scada:
+  time: time
+  id: id
```
ejsimley, line 3: It looks like -25 has a tag name WTUR.TurNam for turbine name. But I'm happy to stay with "id"

openoa/plant.py
```diff
@@ -1633,7 +1678,7 @@ def update_column_names(self, to_original: bool = False) -> None:
 
     def _calculate_turbine_energy(self) -> None:
         energy_col = self.metadata.scada.energy
```
ejsimley, line 1680: as mentioned earlier, we're calling energy WTUR.SupWh in ENTR

examples/data/plant_meta.json
```diff
@@ -9,44 +9,45 @@
...
+    "WROT_BlPthAngVal": "Ba_avg",
+    "WTUR_W": "P_avg",
+    "WMET_EnvTmp": "Ot_avg",
     "time": "Date_time",
-    "wind_direction": "Wa_avg",
-    "windspeed": "Ws_avg"
+    "WMET_HorWdDir": "Wa_avg",
+    "WMET_HorWdSpd": "Ws_avg"
   }
 }
```
ejsimley, line 53: Is there a way to add comments in the json file describing each column name like in the .yml version of plant_meta? I'm guessing there's a reason you left them out here, but just checking!

ejsimley (commented):
Hi Rob, in general this all looks good to me. I left some comments throughout, and pointed out ENTR tag names we're using when there were any differences (I think just for the SCADA energy column right now). 

Also, I'm not aware of -25 tag names for asset columns like rated power, rotor diameter, etc. so I think it's fine to leave those alone for now

ejsimley:
Also, regarding your comment about variable names like energy_corrected, energy_imputed, etc. I'm in favor of leaving these as they are for now and using -25 tag names for "raw" data variables. What do you think?

sphinx/api/types.rst
RHammond2: Thanks, this is really just an expansion of what we were already using to help make that naming switch a little bit easier.

sphinx/api/types.rst
RHammond2: This is updated in the recent wave of commits.

examples/data/cleansed/metadata.yml
RHammond2: Updated.

examples/data/cleansed/metadata.yml
RHammond2: That works for me, we've still got ample time to adopt anything else later on.

examples/data/cleansed/metadata.yml
RHammond2: Updated.

openoa/utils/imputing.py
```diff
@@ -200,12 +200,12 @@ def impute_all_assets_by_correlation(
                 # target_data=data.xs(target_id, level=1).loc[:, [impute_col]],
                 target_data=data.loc[
                     data.index.get_level_values(1) == target_id, [impute_col]
-                ].droplevel("id"),
+                ].droplevel("WTUR_TurNam"),
```
ejsimley, line 203: One potential issue here is that we are assuming imputing will only be performed on turbine data, whereas previously it could have been applied to met tower data. Would it make sense to make the id column an argument that defaults to "WTUR_TurNam". Or we could just use "id" instead of "WTUR_TurNam" in the SCADA data to avoid issues like this.

examples/data/cleansed/metadata.yml
```diff
@@ -0,0 +1,61 @@
+scada:
+  time: time
+  WTUR_TurNam: WTUR_TurNam
+  WTUR_W: WTUR_W
+  WMET_HorWdSpd: WMET_HorWdSpd
+  WMET_HorWdDir: WMET_HorWdDir
+  WTUR_TurSt: WTUR_TurSt
+  WROT_BlPthAngVal: WROT_BlPthAngVal
+  WMET_EnvTmp: WMET_EnvTmp
+  WTUR_SupWh: WTUR_SupWh
```
ejsimley, line 10: Should WTUR_SupWh be removed here, since it gets derived within `PlantData`?

examples/data/plant_meta.yml
```diff
@@ -1,43 +1,45 @@
...
-    windspeed_u: u_50
-    windspeed_v: v_50
+  era5:  # reanalysis product name/ID
+    frequency: H  # timestamp frequency
+    WMETR_EnvPres: surf_pres  # surface pressure, Pa
+    WMETR_EnvTmp: t_2m  # temperature, K
+    time: datetime  # timestamps
+    WMETR_HorWdSpdU: u_100  # u-direction windspeed, m/s
+    WMETR_HorWdSpdV: v_100  # v-direction windspeed, m/s
+    WMETR_HorWdDir: winddirection_deg  # wind direction, degrees
```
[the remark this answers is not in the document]
RHammond2: It's saved because it's already been computed, so for more efficient reloading to `PlantData` and for consistency of results across uses, having the data be the same each time would be ideal. The idea is to save the data that exists, not just to limit it to the columns originally provided or the required columns.

examples/data/cleansed/metadata.yml
RHammond2: From my above comment, the idea is more or less to archive the data that was used so it can be used all over again. In this case, the column will be recomputed because of the decision to make energy internally calculated rather than externally provided.

examples/data/cleansed/metadata.yml
RHammond2: This is used for the examples. I just made it so that if the data is output (as in the 00 example), then the other examples can rely on the prepared data rather than reloading the data and rerunning all the preparation steps. This addition makes it quicker to get an example loaded,  which is nice if someone is running them on their own in Binder or locally.

examples/data/cleansed/metadata.yml
ejsimley: OK, that make sense then

RHammond2:
@jordanperr and @ejsimley i've gone through all the changes again and can't pinpoint anything that would cause any differences in order/randomization/etc. to cause all the assertion errors. My vote is to deal with this stability issue later and get this PR and the next merged in so we can have RC2 complete.

PR #221, opened 2022-12-20
Attach analysis classes to `PlantData`

### LINKED ISSUE #216: Validation of reanalysis data is too limiting
Hi, 

I think the validator for the reanalysis datasets limits them specifically to those called "merra2", "ncep2", "erai" and "era5." I propose removing this requirement, as I wanted to add "era5_monthly" as a dataset, and I'm sure other people will have other datasets too, which might not have these names. In the end I used "erai" to avoid touching the aep.py file!

Thanks,
Charlie

RHammond2:
This PR attaches the analysis classes to `PlantData` at the time of import, maintaining the same behavior as a standalone analysis class import. For example, the following two import patters produce the same results  when used in an analysis.

```python
from openoa import PlantData
from openoa.analysis import WakeLosses

project = PlantData(<kwargs>)

# Original pattern
wake_classic = WakeLosses(project)

# New, equivalent pattern
wake_new = project.WakeLosses()
``` 

This PR also closes #216. @charlie9578 does the proposed solution seem to align with the expectations from our conversation?

@jordanperr, let me know if the implemented pattern works for you. It turns out it was a little more tricky than I originally showed you, but I think it still works nicely because the `openoa` namespace stays highly limited now as well

This should be a fairly straightforward review with the demonstrated pattern implemented in the updated 05 example notebook


charlie9578:
Hi Rob, yes. I've just made a pull request for the Cubico Projects, which uses a "quick fix," but I think what you've done will work for me. Thanks, Charlie

jordanperr:
This pull request implements custom constructors for the analysis classes MonteCarloAEP, EYAGapAnalysis, ElectricalLosses, TurbineLongTermGrossEnergy, and WakeLosses and attaches them as attributes of the PlantData class dynamically using setattr in the init script of the OpenOA module.

I found that the correct docstrings and function signatures can be accessed for a PlantData object in Jupyter Nobteook through the Shift-Tab shortcut. The Control-Space shortcut doesn't work in VSCode's Jupyter Notebook extension. Additionally, the docs are not visible when developing code statically (e.g., in a script, outside of Jupyter).

I think this pull request can be merged as-is, but we must be aware of these limitations.

## In Jupyter:
Screenshot 2023-02-20 at 9 28 22 AM

## In VSCode:
VSCode_nodocs

## Statically, Validate() method supports auto-complete:
Screenshot 2023-02-20 at 9 33 54 AM

## ...  but MonteCarloAEP does not:
Screenshot 2023-02-20 at 9 33 08 AM

jordanperr (approved):
Approve pending review of my comment regarding static code analysis.

ejsimley (approved):
This looks good and worked fine for me when testing it. The only concern I had was whether the option to use the attached analysis classes was documented enough so users would know about this option. I'm wondering if the example in the changelog should be in the docs page somewhere.

PR #226, opened 2023-04-07
Dependency version bump

RHammond2:
This PR is bumping major and minor version of the dependency stack to make this a more modern library.


jordanperr:
I spent a fair bit of time trying to debug the differences we are seeing in the GBM test. I have narrowed it down to the following "minimal" example using only Pandas, Scikit-Learn, Scipy, Numpy, and two data files (derived from La Haute example data): [Download data.zip](https://github.com/NREL/OpenOA/files/11604268/data.zip)

```
from sklearn.ensemble import GradientBoostingRegressor
from sklearn.metrics import mean_squared_error
import numpy as np
import random
import pandas as pd

# Set random seeds
np.random.seed(42)
random.seed(42)

# Load training data
reg_data = np.load("reg_data.npy")
X = reg_data[:, 0:-1]
y = reg_data[:,-1]

# Load prediction data
X_pred = pd.read_csv("reg_inputs_lt.csv")[["era5", "era5_WMETR_EnvTmp"]]

# Fit model
model = GradientBoostingRegressor(n_estimators=370)
model.fit(X,y)

# Predict
y_pred = model.predict(X_pred)

print(y_pred.sum())
```

When I run the above code with Python 3.9.16 X86-64 using scikit-learn==0.24.2 and scipy==1.10.1, I get the following output:

`182.1144285903144`

However, if I run the above code with Python 3.9.16 AARCH-64 using scikit-learn==1.2.2 and scipy==1.10.1, I get the following output:

`182.0842706909885`

I briefly tried to craft an even more minimal example by using random arrays for X,y, and X_pred, but this did *not* produce the same result. As far as I can tell, this bug is dependent on the data.

Issue #228, opened 2023-05-01
Unpin versions of most dependencies


jordanperr:
This issue consolidates issues #178 and #167 

We want to unpin the dependencies in setup.py so that downstream packages (such as the ENTR runtime) don't have to keep downgrading packages to specific versions. The last time this was attempted, we have the following two issues:

- [x] QA Module fails to load due to missing H5PYD dependency, #116 and #220
- [x] Test failure after updating to Pandas 1.3.0 #167 - Tests fail due to small changes in outputs.
- [x] Evaluate new behavior of GradientBoostedModel in scikit-learn > 1.0 and update OpenOA to use new version #178 - Potentially related to #167.
- [x] Unpin versions of all dependencies to latest versions, re-run tests.
- [x] Unpin Python version (right now it's pinned at 3.9)

RHammond2:
@jordanperr I have this in the works, can return to it next week. There are some pretty trivial changes in test results at this point as well, so it doesn't seem particularly worthwhile to do a waterfall analysis of the changes.

RHammond2:
@jordanperr  can you confirmed that this was closed with #226? It appears to be so, but just want your stamp of approval.

jordanperr:
This can be closed. I moved the un-finished part of this to a new issue #239.

PR #234, opened 2023-05-08
Bug fix: Ensure proper handling of errors during partial data definitions

### LINKED ISSUE #233: `PlantData.validate()` fails to validate for missing reanalysis data
In `PlantData.data_validator`, the control flow for when no data is passed to reanalysis is not producing the proper data structure for the final validation method. In [here](https://github.com/NREL/OpenOA/blob/develop_v3/openoa/plant.py#L1260-L1264) there should be a way to generate an empty reanalysis data dictionary with a default data product to ensure the remaining initialization methods run smoothly.

Similarly, this [column validator](https://github.com/NREL/OpenOA/blob/develop_v3/openoa/plant.py#L1458-L1463) is not robust to no reanalysis data being provided by the user.

RHammond2:
This PR fixes a series of small bugs in the validation logic of `PlantData`:
- Ensures that no asset geometry parsing occurs when `PlantData.asset` is `None`
  - Adds `PlantData.calculate_asset_geometries()` for users that add asset data post-validation
- Ensures reanalysis non-definition does not error out the validation steps by adding in a default "product" key for the metadata definition
- Adds a step that ensures all of the reanalysis data products all have matching metadata definitions, or an error occurs early in the initial data loading/validation stage
- Ensures that when `None` is in `PlantData.analysis_type` in addition to other values, that the error raising step is not skipped

Closes #233 

Thank you @jordanperr for identifying that there were a series of related issues when only some data is provided to `PlantData`!


jordanperr:
I validated this PR by reading the code, merging it with [entralliance/OpenOA/from_entr_connector](entralliance:feature/from_entr_constructor)[entralliance:feature/from_entr_constructor](https://github.com/entralliance/OpenOA/tree/feature/from_entr_constructor) and running the electrical losses and AEP notebooks, and also running the plantdata_validator regression tests locally. It looks good to me.

Tests are failing due to known issue https://github.com/NREL/OpenOA/issues/228

PR #237, opened 2023-05-25
Excluding examples module from setup.py packages list

jordanperr:
I discovered that the "examples" module was being installed as a separate package with OpenOA. This pollutes the user's environment and was not expected behavior to me. The culprit is this find_packages() function, which automatically includes all packages in the source tree unless specifically excluded.

Changes:
- Added "examples" to the exclusion list.


RHammond2 (approved):
This makes a lot of sense!

PR #245, opened 2023-07-06
Create SECURITY.md

RHammond2:
This PR adds in the base security and vulnerability protocols to inform users about supported versions of OpenOA.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,18 @@
+# Security Policy
+
+## Supported Versions
+
+Use this section to tell people about which versions of your project are
+currently being supported with security updates.
+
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -0,0 +1,18 @@
+# Security Policy
+
+## Supported Versions
+
+Use this section to tell people about which versions of your project are
+currently being supported with security updates.
```
ejsimley, line 6: This statement sounds like it's an instruction for creating the security policy. Consider changing to "Versions currently being supported with security updates are listed below."

ejsimley (approved):
This is a nice addition. I just have a comment on the language used.

Issue #214, opened 2022-10-13
Documentation Improvements


RHammond2:
This issue is a reflection of the unaddressed content improvements from @jordanperr and @ejsimley in #207.

- [x] Add more user-friendly information to the Overview section of `sphinx/index.rst`
- [x] Introduce more of the code in the Overview section of `sphinx/index.rst` (@ejsimley)
  - The Overview section would be a good place to describe all of the analysis methods at a high level and the utils toolkits as well, so that the "Plant Data", "Analysis Methods", and "Utils" pages on the top menu make more sense. This way users can understand what OpenOA can be used for. I can work on this. And then the order of the pages in the table of contents/menu might make more sense as plant data, analysis methods, utils, then examples since Plant Data is at the core of all things OpenOA, then analysis methods are the most important high-level methods, followed by utils, with examples I think making sense after the software components are explained.
- [x] Improve the readability of the installation and contributing sections
- [x] Provide a better introduction to the examples in `sphinx/examples/index.rst`
- [x] The reanalysis section of `sphinx/api/types.rst` needs to be rewritten to be more sensible and applicable to users
- [x] Rethink the section titling for the `sphinx/api/analysis.rst` and if we want more description in the RST file for each of the analysis categories (or methods) or just to modify the docstring content that populates these pages
- [x] Link to our analysis papers that helped develop these methods or utilized these methods (should we feature other papers too?--this could be a good way to engage the community and highlight their work that uses the code @ejsimley or @jordanperr or @Dynorat thoughts on this part would be good)

ejsimley:
When looking at the new documentation I noticed that example 2b is not included in the examples documentation. Should this be added, or is there a reason it isn't included yet?

RHammond2:
It was intentional but only because it hadn't been  updated with example 2, so I was unsure if it should be included or not. Happy to bring up to date and reinstate it though.

jordanperr:
Additionally, it would be nice to have some tables or lists in the documentation showing the tag names, units, frequencies, and any other requirements, broken out by analysis type.

We could also compile a superset of all possible columns into one large schema document.

I suggest this all be defined in JSON files somewhere (such as an Analysis_Requirements JSON file, and a PlantSchema JSON file), which is then read in by Sphinx to generate the tables automatically. Related to #227

PR #246, opened 2023-07-10
Update Getting Started Instructions and Settings

RHammond2:
This PR addresses the third bullet point in #214 to improve upon the legibility of the getting started, contributing, and installation instructions. This is accomplished by a basic reorganization of the content to be more streamlined and by communicating the ideas more clearly and inline with using Python in 2023.

This also restructures the way that the examples are displayed so that can each be their own page on the documentation site instead of a single page with a high level of nesting.


contributing.md
```diff
@@ -1,57 +1,148 @@
...
+   will only check files that have been modified and staged (`git add <your-changed-file>`).
+
+   ```bash
+   pre-commit install
+   ```
+
+## Keeping your fork in sync with NREL/OpenOA
+
+The "main" OpenOA repository is regularly updated with ongoing research at NREL and beyond. After
+creating and clonging your fork from the previous section, you might be wondering how to keep it
```
jordanperr, line 45: clonging

contributing.md
```diff
@@ -61,46 +152,133 @@ lines, etc.), ``isort`` for automatic import sorting, and ``flake8`` for linting
...
+### Documentation
+
+Include any relevant changes to inline documentation, docstrings, and any of the RST files located
+in `OpenOA/sphinx/`. Pull requests will not be accepted until these changes are complete.
+
+### Changelog
+
+All changes must be documented appropriately in CHANGELOG.md in the [Unreleased] section.
 
 ## Release Process
```
jordanperr, line 257: May be worth adding a couple sentences clarifying this section is for maintainers of the NREL/OpenOA repo only, and does not apply to your own fork.

contributing.md
```diff
@@ -1,57 +1,148 @@
 
-Contributing
-============
+# Contributing
+
+
+## Getting Started
+
+All users wishing to contribute to OpenOA, or modify the code in any way, even for personal use,
+should follow much of the below guides.
```
jordanperr, line 8: I suggest rephrasing this to be more declarative and less instructive:

"These contributing guidelines should be read by software developers wishing to contribute code or documentation changes into OpenOA, or to push changes upstream to the main NREL/OpenOA repository."

jordanperr (changes requested):
- [x] When you click on "Getting Started" or "[Installation, Using, and Contributing](https://openoa.readthedocs.io/en/develop_v3/getting_started/index.html#installation-using-and-contributing)" from the homepage, it goes to an index of the Installation instructions, which then requires another click to actually go to the installation guide. By contrast, the Changelog is printed right there on the page. This is confusing to me.
- [x] The "Overview" section of the examples lists some bullets about each example, and then produces a large table of contents below these bullets with links to the examples. Is it possible to add links to each examples up in the bullet list in the overview? In addition, the bullets do not mirror the titles of each example. (Electrical Losses, for example, is called "[Gap Analysis Step 3: Estimate Electrical Losses Using SCADA and Revenue Meter Data](https://openoa--246.org.readthedocs.build/en/246/examples/04_electrical_losses.html)" in the TOC.  https://openoa--246.org.readthedocs.build/en/246/examples/index.html
- [ ] Pandas tabular output from example notebooks is not formatted correctly in Safari. It works fine in Firefox so I'm inclined to just ignore for now: https://openoa--246.org.readthedocs.build/en/246/examples/02_plant_aep_analysis.html#Step-3:-Process-the-data-into-monthly-averages-and-sums
- [ ] "[Gap Analysis Step 1: Estimate the AEP and Its Uncertainty](https://openoa--246.org.readthedocs.build/en/246/examples/02_plant_aep_analysis.html)" is printed twice in the Examples TOC: https://openoa--246.org.readthedocs.build/en/246/examples/index.html#table-of-contents

RHammond2:
@jordanperr, below are some replies as they relate to your bullets above, but separated for ease of cross referencing the original point.
- The installation instructions sit one step lower than the changelog because there are many considerations in terms of what a user might want from them, whether that's a basic user or someone contributing back to the repo. The chagelog, from my perspective also offers a lot of helpful information for getting started, particularly as we significantly update the repository. Additionally, the installation instructions having an extra click to get to the `pip install` prompt is largely in line with what many other code bases do that have varying options for what or how a user might want to install the package.
- At least when I build locally, it doesn't like to link to the actual example for some reason, and I think that has to do with the existing solutions to referencing examples that are contained outside of the documentation folder. The only way I can think of for getting around this issue is to hard code the link, but that will cause problems for others trying to access the different branches on the docs site because it will be hard-coded to a single branch's reference page. I'll update the naming convention as well, so it more clearly tracks the notebook naming.
- This tracks with Firefox as well, particularly for how it displays on the RTD vs a local build, which didn't have issues for me. I'll look into this, but I assume there is a setting that needs to be updated for a version change to properly allow for sideways scrolling.
- Good catch, this seems to be coming from our, not at all updated example modified-AEP example. I'll just update that in this PR per #247.

sphinx/getting_started/install.rst
```diff
@@ -13,5 +13,69 @@
 Install
 #######
 
-.. include:: ../../readme.md
-   :parser: myst_parser.sphinx_
+Pip
```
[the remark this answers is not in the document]
RHammond2: It seemed like a lot of noise to add the whole README to the installation instructions, especially now that the actual installation and contributing content is broken down a bit differently. It's technically more work to keep both up  to date, but I don't anticipate having any changes to the installation instructions that make updating two sources a real concern. Now that there will also be a contributors section in the README, I don't  think that needs to get brought into the installation documentation page.

openoa/analysis/aep.py
```diff
@@ -1003,7 +1003,12 @@ def run_AEP_monte_carlo(self):
             iav[n] = gross_lt_annual.std() / gross_lt_annual.mean()
             avail_pct[n] = avail_lt_losses
             curt_pct[n] = curt_lt_losses
-            lt_por_ratio[n] = (gross_lt.sum() / self._run.num_years_windiness) / gross_por.sum()
+            gps = (
+                gross_por.sum()
+                if not isinstance(gross_por, (pd.Series, pd.DataFrame))
+                else gross_por.values.sum()
+            )
+            lt_por_ratio[n] = (gross_lt.sum() / self._run.num_years_windiness) / gps
```
jordanperr, line 1011: What is the purpose of this code change? This is not in develop_v3. Just checking to make sure it's not a mistake.

openoa/utils/machine_learning_setup.py
```diff
@@ -190,7 +190,7 @@ def hyper_optimize(
             param_distributions=self.hyper_range,
             n_iter=n_iter_search,
             scoring=self.my_scorer,
-            verbose=1,
+            verbose=0,
```
jordanperr, line 193: Would be preferable to be able to set this globally, or pass in the verbosity level.

jordanperr (approved):

openoa/utils/plot.py
```diff
@@ -2118,7 +2118,7 @@ def plot_yaw_misalignment(
                 ],
                 color=curve_fit_color_code,
                 linestyle="--",
-                label=f"Max. Power Vane Angle = {round(curve_fit_params_ws[:,i,1].mean(),1)}$^\circ$",
+                label=f"Max. Power Vane Angle = {round(curve_fit_params_ws[:,i,1].mean(),1)}$^\circ$",  # noqa: W605
```
ejsimley, line 2121: nice that you found a fix for this! (assuming this resolves the error that gets thrown during the pre-commit workflow)

RHammond2: That's exactly what this does!

openoa/utils/machine_learning_setup.py
RHammond2: I agree, my one concern is that by using tqdm for tracking iterations alongside anything more granular than 0 produces the massive output that we had issues with in past versions. I'm going to add this to the logging issue unless there's a reason we should address it here and now.

openoa/analysis/aep.py
RHammond2: This has to do with how we single value Series can be interacted with vs numpy arrays. Essentially in a future version of pandas the original version will just break, which makes for a really annoying warning coming out of pandas.

ejsimley (approved):
The documentation looks great! I just have a few minor comments to consider.

readme.md
```diff
@@ -43,59 +43,71 @@ If you use this software in your work, please cite our JOSS article with the fol
...
 ```
 
 - In Windows, an ImportError regarding win32api can also occur. This can be resolved by fixing the version of pywin32 as follows:
 
-```
+```bash
 pip install --upgrade pywin32==255
 ```
 
 ### Development
```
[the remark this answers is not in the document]
RHammond2: Good idea, I linked to the place where it will be once these updates are live and in v3 so we don't have to change it again, so the documentation site link will break if you try to go to it

setup.py
```diff
@@ -50,10 +51,16 @@
...
         "flake8-docstrings",
-        "pytest",
-        "pytest-cov",
+    ],
+    "examples": [
+        "jupyterlab",
+        "xarray",
+        "dask",
+        "netcdf4",
+        "cdsapi",
```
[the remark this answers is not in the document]
RHammond2: That's a good point, I'd been starting to ponder whether they should all just be a part of the requirements, and I think this is what I needed. We also use ipython in the code now for the updated PlantData repr method, so I think it would just be easier to be more inclusive all around like we did when updating for the WTK dependency.

readme.md
```diff
@@ -104,53 +116,64 @@ git add .pre-commit-config.yaml
```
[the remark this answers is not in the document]
RHammond2: That seems like a really obvious and good improvement, done!

Issue #230, opened 2023-05-01
Merge in Static Yaw Misalignment analysis from ENTR Fork


jordanperr:
Source: https://github.com/entralliance/OpenOA/tree/feature/static_yaw_misalignment
Destination: https://github.com/NREL/OpenOA/tree/develop_v3

Without entr connector

- [ ] Scientific regression tests should be added and pass using NREL/OpenOA github actions.
- [ ] Code review by Rob and/or myself.

PR #249, opened 2023-07-12
Static yaw misalignment analysis method

ejsimley:
This pull request contributes a new analysis method for estimating static yaw misalignment for individual turbines using SCADA data. The PR includes the following changes and additions:

- The `yaw_misalignment` analysis module which contains the `StaticYawMisalignment` class for performing the yaw misalignment assessment using a `PlantData` object
- A new `plot_yaw_misalignment` function in the `plot` utils module, which is used by the `StaticYawMisalignment` class to help visualize yaw misalignment
- The notebook "07_static_yaw_misalignment.ipynb", which illustrates the yaw misalignment method applied to the LHB data
- Updates to the sphinx documentation to include the yaw misalignment analysis method 
- Yaw misalignment regressions tests
- Updated analysis requirements in `metadata.py`, including the addition of asset data requirements for the `WakeLosses` and `StaticYaw Misalignment` analysis methods 

This pull request closes issue #230


openoa/analysis/yaw_misalignment.py
```diff
@@ -0,0 +1,571 @@
...
+    wind turbine. If :py:attr:`UQ` is selected, the following steps are performed multiple times
+    using Monte Carlo simulation to produce a distribution of static yaw misalignment estimates
+    from which 95% confidence intervals can be derived:
+      1. Timestamps containing power curve outliers are removed. Specifically, pitch angles are
+         limited to a specified threshold to remove timestamps when turbines are operating in or
+         near above-rated conditions where yaw misalignment has little impact on power performance.
+         Next to increase the likelihood that power performance deviations are caused by yaw
+         misalignment, a power curve outlier detection filter is used to remove timestamps when the
+         turbine is oeprating abnormally. If :py:attr:`UQ` is selected, power curve outlier
+         detection parameters will be chosen randomly each Monte Carlo iteration.
```
RHammond2, line 67: "for each"

openoa/analysis/yaw_misalignment.py
```diff
@@ -0,0 +1,571 @@
...
+
+        if self._use_power_coeff:
+            power_performance_label = "Normalized Cp (-)"
+        else:
+            power_performance_label = "Normalized Power (-)"
+
+        if turbine_ids is None:
+            turbine_ids = self.turbine_ids
+        else:
+            if (set(turbine_ids) - set(self.turbine_ids)) != set():
```
RHammond2, line 526: Pretty minor, but this can be done with Python's set math as follows:

```python
if (set(turbine_ids).difference(self.turbine_ids)):
```

openoa/utils/plot.py
```diff
@@ -1932,3 +1932,284 @@ def plot_wake_losses(
...
+            `ax.plot()` for plotting lines for the power performance vs. wind vane plots. Defaults to {}.
+        plot_kwargs_line (:obj:`dict`, optional): Additional plotting keyword arguments that are passed to
+            `ax.plot()` for plotting vertical lines indicating mean vane angle and vane angle where
+            power is maximized. Defaults to {}.
+        plot_kwargs_fill (:obj:`dict`, optional): If `UQ` is True, additional plotting keyword arguments
+            that are passed to `ax.fill_between()` for plotting shading regions for 95% confidence
+            intervals for power performance vs. wind vane. Defaults to {}.
+        legend_kwargs (:obj:`dict`, optional): Additional legend keyword arguments that are passed to
+            `ax.legend()` for the power performance vs. wind vane plots. Defaults to {}.
+    Returns:
```
RHammond2, line 2005: I think there just needs to be an extra line before "Returns".

openoa/utils/plot.py
```diff
@@ -1932,3 +1932,284 @@ def plot_wake_losses(
...
+        None | tuple[matplotlib.pyplot.Figure, matplotlib.pyplot.Axes]:
+            If `return_fig` is True, then the figure and axes object(s) corresponding to the yaw
+            misalignment plots are returned for further tinkering/saving.
+    """
+
+    from openoa.analysis.yaw_misalignment import cos_curve
```
RHammond2, line 2011: I think it might be best to have this method in the `utils` subpackage, but I'm not totally sure where it belongs in there.

openoa/utils/plot.py
```diff
@@ -1932,3 +1932,284 @@ def plot_wake_losses(
...
+                    0.01 * np.floor(y_min / norm_factor / 0.01),
+                    0.01 * np.ceil(y_max / norm_factor / 0.01),
+                ],
+                color=curve_fit_color_code,
+                linestyle="--",
+                label=f"Max. Power Vane Angle = {round(curve_fit_params_ws[:,i,1].mean(),1)}$^\circ$",
+            )
+
+            axs[int(np.floor(i / N_col))][i % N_col].set_title(
+                f"{ws} m/s\nYaw Misalignment = {np.round(np.mean(yaw_misalignment_ws[:,i]),1)}$^\circ$ [{np.round(np.percentile(yaw_misalignment_ws[:,i],2.5),1)}$^\circ$, {np.round(np.percentile(yaw_misalignment_ws[:,i],97.5),1)}$^\circ$]"
```
RHammond2, line 2121: I'm not normally one to make single-use variables, but I think it might be best to compute the numbers that go in the f-string just above this line because it's so long. Additionally, parentheses can be used to write strings across multiple lines like so:

```python
ax.set_title(
    (
        f"{ws} m/s\nYaw Misalignment = "
        f"{np.round(np.mean(yaw_misalignment_ws[:,i]),1)}$^\circ$ "
        f"[{np.round(np.percentile(yaw_misalignment_ws[:,i],2.5),1)}$^\circ$, "
        f"{np.round(np.percentile(yaw_misalignment_ws[:,i],97.5),1)}$^\circ$]"
    )
)
```

openoa/utils/plot.py
```diff
@@ -1932,3 +1932,284 @@ def plot_wake_losses(
...
+        N_col = 3
+
+    for i, ws in enumerate(ws_bins):
+        if UQ:
+            norm_factor = curve_fit_params_ws[:, i, 0].mean()
+
+            y_min = np.nanmin(np.nanpercentile(power_values_vane_ws[:, i, :], 2.5, 0))
+            y_max = np.nanmax(np.nanpercentile(power_values_vane_ws[:, i, :], 97.5, 0))
+
+            axs[int(np.floor(i / N_col))][i % N_col].fill_between(
```
RHammond2, line 2085: It seems like it'd be best if the axis in `axs` could be pulled at the top of the for loop to reduce the number of times this index is calculated to help make it easier to read. Typically, I prefer something like `ax = axs[int(np.floor(i / N_col))][i % N_col]`, so each subsequent call can just be `ax.plot()`.

openoa/utils/plot.py
```diff
@@ -1932,3 +1932,284 @@ def plot_wake_losses(
...
+            )
+
+    for i in range(N_col):
+        axs[int(np.floor((len(ws_bins) - 1) / N_col))][i].set_xlabel("Wind Vane Angle ($^\circ$)")
+
+    mean_yaw_misalignment = np.mean(yaw_misalignment_ws)
+    if UQ:
+        yaw_misalignment_95CI = np.percentile(np.mean(yaw_misalignment_ws, 1), [2.5, 97.5])
+        fig.suptitle(
+            f"Turbine {turbine_id}, Yaw Misalignment = {np.round(mean_yaw_misalignment,1)}$^\circ$ [{np.round(yaw_misalignment_95CI[0],1)}$^\circ$, {np.round(yaw_misalignment_95CI[1],1)}$^\circ$]"
```
RHammond2, line 2199: Same here, particularly, the `np.round()` can be done when `yaw_misalignment_95CI` is created. You can also feel free to shorten the variable name if you like, even just dropping "yaw" or "misalignment"

RHammond2 (changes requested):
This is really good, thanks for yet another analysis class, @ejsimley! I've made a few small comments and amendments (only for really minor points about speedups via NumPy, or lower-level setup pieces that were missing), but nothing that fundamentally changes the work you've done.

I'll look into the failing tests a little bit more and the notebook commentary in more detail in another comment, but wanted to get you feedback on the code contributions first.

RHammond2:
@ejsimley, below I've made some comments on the new example notebook. Overall, it's a really great demonstration of the capabilities, nice work!

#### Cell 5

The plots should be generated by `openoa.utils.plot.plot_by_id(project.scada, "asset_id", "WMET_HorWdSpd", "WROT_BlPthAngVal"). However, I realize this also leaves room for improvement in that method for the actual axis labesl being additional input, in place of relying on the variable names themselves, and so I just created Issue #251.

#### Yaw misalignment detection without uncertainty quantification

I'd probably just have all "(`UQ=False`) instead of "(`UQ`=False)"

#### Cell 10

Looks like the empty axes are still being shown, so it'd be good to have them be removed if they're not in use. Let me know if you want any help with that.

Is your suggestion of power-weighted misalignment in the works? That sounds pretty interesting as an alternative, and if not, could be slightly misleading to the reader.

#### Conclusion

Should there be a sentence or two to end the notebook on what do with any of this? It feels like it ends abruptly

openoa/analysis/yaw_misalignment.py
```diff
@@ -0,0 +1,571 @@
...
+from openoa.logging import logging, logged_method_call
+
+
+logger = logging.getLogger(__name__)
+NDArrayFloat = npt.NDArray[np.float64]
+plot.set_styling()
+
+
+def cos_curve(x, A, Offset, cos_exp):
+    # Cosine exponent curve as a function of yaw misalignment for curve fitting
```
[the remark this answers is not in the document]
ejsimley: Updated docstring

openoa/analysis/yaw_misalignment.py
```diff
@@ -0,0 +1,571 @@
...
+    return A * np.cos((np.pi / 180) * (x - Offset)) ** cos_exp
+
+
+@define(auto_attribs=True)
+class StaticYawMisalignment(FromDictMixin):
+    """
+    A method for estimating static yaw misalignment for different wind speed bins for each specified wind turbine as well as the average static yaw msialignment over all wind speed bins using turbine-level SCADA data.
```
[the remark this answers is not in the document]
ejsimley: Revised

openoa/analysis/yaw_misalignment.py
```diff
@@ -0,0 +1,571 @@
...
+
+    @plant.validator
+    def validate_plant_ready_for_anylsis(
+        self, attribute: attrs.Attribute, value: PlantData
+    ) -> None:
+        """Validates that the value has been validated for a static yaw misalignment analysis."""
+        if set(("StaticYawMisalignment", "all")).intersection(value.analysis_type) == set():
+            raise TypeError(
+                "The input to 'plant' must be validated for at least 'StaticYawMisalignment'"
+            )
```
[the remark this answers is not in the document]
ejsimley: I followed the example in your link and moved the validation steps to `__attrs_post_init__()`. Take a look and see if this is what you're looking for.

openoa/analysis/yaw_misalignment.py
```diff
@@ -0,0 +1,571 @@
...
+            self._max_power_filter = (0.92, 0.98)
+        else:
+            self._max_power_filter = 0.95
+
+        if power_bin_mad_thresh is not None:
+            self._power_bin_mad_thresh = power_bin_mad_thresh
+        elif self.UQ:
+            self._power_bin_mad_thresh = (4.0, 10.0)
+        else:
+            self._power_bin_mad_thresh = 7.0
```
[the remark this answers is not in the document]
ejsimley: As we discussed, I have not tried to address this yet

openoa/analysis/yaw_misalignment.py
```diff
@@ -0,0 +1,571 @@
...
+        self._df_turb = self._df_turb.loc[self._df_turb["WROT_BlPthAngVal"] <= self._pitch_thresh]
+
+        # Apply bin-based filter to flag samples for which wind speed is greater than a threshold from the median
+        # wind speed in each power bin
+        turb_capac = self.plant.asset.loc[turbine_id, "rated_power"]
+        bin_width_frac = 0.04 * (self._run.max_power_filter - 0.01)
+        flag_bin = filters.bin_filter(
+            bin_col=self._df_turb["WTUR_W"],
+            value_col=self._df_turb["WMET_HorWdSpd"],
+            bin_width=bin_width_frac * turb_capac,
```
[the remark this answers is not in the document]
ejsimley: I think it makes sense for them to be customizable. I added two new arguments to the `run` method so those two filtering parameters can be specified: `num_power_bins` and `min_power_filter`.

openoa/analysis/yaw_misalignment.py
ejsimley: Updated

openoa/utils/plot.py
ejsimley: As we discussed, I left this alone for now.

openoa/utils/plot.py
ejsimley: Good suggestion. This has been updated.

openoa/utils/plot.py
ejsimley: Updated

ejsimley:
Hi @RHammond2, thanks for the careful review and for helping to improve the code! I addressed all of the comments except for the ones that we discussed adding as separate issues. The unaddressed comments are:

- `yaw_misalignment.py`: It seems like we should have an input validation method similar to the `_analysis_validators.validate_UQ_inputs()` for the run method inputs. I can make a separate PR to address this more effectively because I think much of the other analysis methods' inputs should be inputs to `run()` and not the class itself.
- `plot.py`: I think it might be best to have this method in the utils subpackage, but I'm not totally sure where it belongs in there.
- The plots should be generated by `openoa.utils.plot.plot_by_id(project.scada, "asset_id", "WMET_HorWdSpd", "WROT_BlPthAngVal")`. However, I realize this also leaves room for improvement in that method for the actual axis labesl being additional input, in place of relying on the variable names themselves, and so I just created Issue #251 

Let me know if you'd like me to work on any of those, or if you have any more comments on this PR. 

Also, there are still a couple minor differences causing the tests to fail. I'm not sure what is the best way to address this.

openoa/analysis/yaw_misalignment.py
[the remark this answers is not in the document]
RHammond2: I have this in #252

openoa/analysis/yaw_misalignment.py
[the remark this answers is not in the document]
RHammond2: I'm not seeing this reflected, do you need to push the changes still?

RHammond2:
Thanks @ejsimley! There seems to be one missing update, which is the only unresolved comment, otherwise this is ready to go from a code perspective.

As for the failing test, it seems that it's back to just the one invalid result in `test_yaw_misaliginment_with_UQ` for the `expected_yaw_mis_results_95ci_ws` values. I'm not too sure why this would be the case because they pass on my end still. Do they work on yours by any chance? I looked into the installed dependency versions previously, and didn't come up with anything on that front like we had previously. I wonder if we should tag Jordan in at this point, though I'll be out for the next week and can think a little more about it when I'm back.

openoa/analysis/yaw_misalignment.py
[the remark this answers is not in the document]
ejsimley: Hi Rob, I checked and those changes are in the code. In addition to the arguments to the `run` method I mentioned above, the `bin_min` argument to the bin filter function uses the `self._min_power_filter` value instead of the hardcoded value of 0.01 used previously. And the `bin_width_frac` variable used in the `bin_width` argument is based on `self._num_power_bins`, which is an argument to `run` instead of hardcoded to 25. Let me know if I'm missing something, or perhaps I misunderstood your original comment.

ejsimley:
Hi @RHammond2, thanks for reviewing the changes! See my response about the unresolved comment, which I think is already incorporated. 

Regarding the tests, they all pass on my laptop as well. I don't have any ideas about how to address the lone test failure on the CI pipeline (other than further reducing the test precision), so we should discuss how to proceed.

openoa/analysis/yaw_misalignment.py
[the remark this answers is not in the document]
RHammond2: Ok, I think I didn't read closely enough, and looked for the wrong change. This is great, thank you!

RHammond2 (approved):

PR #250, opened 2023-07-18
Utils Methods and TIE Speedups

RHammond2:
This PR address the remaining, easy-to-address slowness that bogs down the TIE analysis method with Example 3 run time decreasing to ~2.5-3.5 minutes from the previous ~16 minute runtime. The speedups come from:
1. converting operations relying Pandas `Series` methods to the underlying NumPy methods and converting the data to NumPy arrays
2. Converting `openoa.utils.filters.bin_filter` to array math from nested `for` loops on top of the improvements from 1

One of the tests was changed because the input data was using mixed data types, but all data being passed through the the various `utils` methods will be converted to their correct data types, or should by the time they're used in an analysis. In either case mixing string and float data made little sense to test the robustness of a method operating on floats.


ejsimley (approved):
Hi @RHammond2, thanks for making these improvements. I ran the example 03 notebook and the speedup from the new filter implementation is huge! My only request is to run the last few cells of the example 03 notebook so the outputs are already printed when viewing the notebook.

PR #255, opened 2023-09-27
Final Cleanup for v3

### LINKED ISSUE #125: Remove hard-coded constants
In response to JOSS review, we should place a code review on the roadmap to remove any hard-coded constants that may have crept in over the years.

> I would also caution against too much hardcoding within the code. In some instances, you could make better use of dependency functionality (e.g. operational_analysis/toolkits/unit_conversion.py:convert_power_to_energy could just resample the dataframe rather than expecting one of five hardcoded input time frequencies). In other cases, 'translations' could be kept inside e.g. YAML files, so that they're all in one place, configurable and out of the way of the Python code (e.g. scada_map in examples/project_ENGIE.py).

### LINKED ISSUE #214: Documentation Improvements

### LINKED ISSUE #229: Programmatically generate requirements for analysis
We would like to be able to use the analysis_requirements to generate a query for the ENTR warehouse. This works now if you're only using required columns, but becomes tricky when you want to use optional columns. The problem is that the use of optional columns are dictated by the arguments provided to the analysis, and these may not be known at the time the PlantData is created. I propose adding an "Analysis.generate_schema()" method to analysis classes, which generates the requirements dictionary on-the-fly, based on the arguments provided to init.

Old Workflow:

```
plant = PlantData.from_entr("La Haute Borne", analysis_type="MonteCarloAEP")
pa = MonteCarloAEP(plant, arg1="foo", arg2="bar")
pa.run(arg3="baz")
```

Proposed Workflow:

```
pa = MonteCarloAEP(arg1="foo", arg2="bar")
plant = PlantData.from_entr("La Haute Borne", schema=pa.get_schema())
pa.run(plant, arg3="baz").
```

RHammond2:
This is the final cleanup for take care of the last remaining threads in the issues.

Closes #125 
- Any remaining constant is not one that should ever be changed, but is created at the top of the module if it's commonly used.

Closes #229 
- There is not a formal means through which a schema is created, but all of the information required for a user to port their data to a compatible format and schema is abundantly documented.

Closes #214
- Much of the descriptions have been revisited to offer minor clarifications and reorganization to read more clearly, though more work could likely be done in the FY24/v3.1 scope of work, and should be considered there as a more specific listing of improvements.


openoa/analysis/wake_losses.py
```diff
@@ -465,6 +481,10 @@ def run(
             self.assume_no_wakes_high_ws_LT_corr = assume_no_wakes_high_ws_LT_corr
         if no_wakes_ws_thresh_LT_corr is not None:
             self.no_wakes_ws_thresh_LT_corr = no_wakes_ws_thresh_LT_corr
+        if min_ws_bin_lin_reg is not None:
+            self.min_ws_bin_lin_reg = min_ws_bin_lin_reg
+        if bin_count_thresh_lin_reg is not None:
+            self.bin_count_thresh_lin_reg = bin_count_thresh_lin_reg
```
jordanperr, line 487: Why not just define the default value for min_ws_bin_lin_reg and bin_count_thresh_lin_reg as default arguments in the run() argument list?

I see two problems with the current implementation:
1. Added complexity (lines 484-487, the use of an attrs field)
2. Unless I'm looking at this wrong - arguments to run() will now "stick" - and become the new default value next time run() is called. This is inconsistent with the documentation which reads "Defaults to 3.0." In fact, it will default to 3.0 or to whichever value was last passed to "run()".

openoa/plant.py
```diff
@@ -1068,23 +1066,36 @@ def parse_asset_geometry(
...
         Returns: None
             Sets the asset "geometry" column.
         """
+        # Check for metadata inputs
+        if utm_zone is None:
+            utm_zone = self.metadata.utm_zone
+        if reference_longitude is None:
+            reference_longitude = self.metadata.reference_longitude
+        if reference_system is None:
+            reference_system = self.metadata.reference_system
```
jordanperr, line 1089: I think we have a similar thing going on here - except the reference default reference system will be changed for the whole PlantData (and, hypothetically, any other PlantData which shares the same Metadata)

jordanperr (changes requested):
- Potential issue with "sticky" default value.
- I don't think this closes #229, but rather moves this idea to the backlog.

openoa/plant.py
RHammond2: This is data that shouldn't actually changed given it's the reference data, but I could see not resetting it if that's a real hangup.

openoa/analysis/wake_losses.py
RHammond2: So I'm going to push back on this because this relates back to an implementation schematic that's been lingering without feedback for a couple of months. In the new API we're now using the `self.parameter` when accessing instead of just `parameter` as we had done previously. Additionally, we're using a set of attrs validators to check the inputs in many cases to ensure they're valid, so we'd have to then revert back to just manually checking all the inputs.

Would it be helpful to add in a method like the following to allow a resetting of the values when a user wants to clear the slate?

```python
def reset_defaults(self, which: str | list[str] | tuple[str], None = None):
    """"Reset all or a subset of the analysis parameters back to their defaults."""
    # Define the analysis class run parameters
    valid = ["""analysis class run parameters"""]
    
    # If None, set to all run parameterizations
    if which is None:
        which = valid
    
    # Check that all values of which are valid
    ...

    # Reset the selected values back to their defaults
    for attribute in attrs.fields(self):
        if attribute.name in which:
            setattr(self, attribute.name, attribute.default)
```

RHammond2:
@jordanperr  for #229 I think for me the ask is still a bit vague and not remotely possible if an analysis class requires a `PlantData` object for initialization, so if you could provide more details on the what then we might be able to get something in there, but as-is I would close it and request a new issue with more specifics. Similarly, the schema are provide through `openoa.schema.metadata.ANALYSIS_REQUIREMENTS`, so a user could just import that and use it as needed.

However, if we're thinking of enabling bringing a custom validation set, then that's maybe a separate idea, though, and would enable additional workflows like a user-defined analysis class/method, which is pretty useful for more customized analysis workflows.

Issue #260, opened 2023-11-30
Remove jupyterhub requirement


jordanperr:
IPython has a module-level import in plant.py, which causes jupyterhub to be a package-level requirement. Jupyterhub is a very heavy install and may conflict with commercial jupyter-like environnments.

- Remove jupyterhub as a requirement
- Remove module-level imports of ipython
- Move all such imports into guarded sections that check for ipython.

Issue #259, opened 2023-11-30
Review package requirements in setup.py


jordanperr:
The requirements in setup.py have become out of date and have too many extra dependencies in the core requrirements.

Some clear places to start:
- pyspark, not required anymore
- jupyterlab, review if still needed or move to an extra
- statsmodels is in there twice
- dask, review if still required

- Hex is pinned to numpy 0.23, but we require 0.24. Is there a reason to require 0.24?

RHammond2:
Here is the complete roundup from our discussion last week:

- [ ] update `attrs` to `attrs>=22.2`
- [ ] add `ipython` as a first class dependency
- [ ] remove duplicated `statsmodel`
- [ ] remove `pyspark`
- [ ] remove `examples/requirements.txt`
- Add the following install options:
  - [ ] add a note to the appropriate sections in the documentation indicating these updates
  - [ ] examples
    - `jupyterlab`
  - [ ] reanalysis
    - `cdsapi`
    - `xarray`
    - `netcdf4`
    - [ ] Add a note to the Cubico example stating this option will need to be installed first
  - [ ] nrel-wind (or similar naming)
    - `h5pyd` (wind toolkit)

RHammond2:
It turns out that Dask is required for xarray to load in the reanalysis data, so I've updated xarray to be `xarray[parallel]` to be up to xarrays spec on that front, rather than managing it separately.

RHammond2:
This was addressed in #262, and once merged into main, will be closed.

PR #262, opened 2023-12-07
Patch: Dependency updates

RHammond2:
This PR addresses #259 and #260 to ensure the base OpenOA installation is not bogged down by Jupyter for users that will not require it, and to reassess the import requirements modifiers. Additionally, I've shifted some of the configuration files contained in the top-level of `OpenOA/` to be encapsulated in pyproject.toml to reduce some of the visual clutter as part of this process. Finally, I added pytest flags for `--unit` and `--regression` to just tell pytest which tests to run as a convenience.

Something to be addressed later is that our pytest coverage modifiers don't seem to actually work, which I think is ok because it's more honest that way, but I'm interested to see if that's something we should just remove wholesale.


jordanperr:
I downloaded this and was able to install OpenOA and run the tests successfully.

The code coverage worked fine for me. Using the command: `pytest --unit --cov=openoa` I see a table printed to to standard out containing computed code coverage.

The addition of pyproject.toml seems reasonable to me. It now seems there is some redundancy between the pyproject.toml and the setup.py file. We could inspect the possibility of migrating things from setup.py to pyproject.toml, outside the scope of the PR.

I also ran into some other issues with the example notebooks, which are outside the scope of this pull request, and placed them in a new issues: #263 #264 

Comments which I'd like to discuss:

- [ ] While running the example notebooks, I noticed an Import error for h5pyd in 00_intro so I added this to the "examples" extra.
- [ ] The cubico notebook requires the reanalysis module (which is highlighted at the top of the notebook). I think it would be better to simply include cdsapi and xarray in the examples extra instead of having users install the reanalysis package.
- [ ] Overall, there are now a large number of dependency extras, which could be confusing to users. What if instead we just have three extras: `all-deps` (which will include all non-develop dependencies), `tests` (pytest and pytest-cov), and `develop` (tests + linting)?

RHammond2:
@jordanperr thanks for the review! I think it makes a lot of sense to have the examples contain the `nrel-wind` and `reanalysis` modifiers to ensure the examples will run regardless of which are intended to be used.

We also still don't support Python 3.11 because a few of our dependencies don't support it as the last time I checked (June?). I'm happy to revisit this, but I'll likely be rolling back the 3.11 support statement.

The TOML usage was introduced somewhere in the v3 work, but for whatever reason I didn't fully migrate all the non-setup settings. That said, I was definitely holding off converting the the setup.py to a pyproject.toml until a later point in time as it involves changing how we manage our versioning.

Based on your comments, I'd recommend the following dependency modifiers:
- `reanalysis`: leave as-is
- `nrel-wind`: leave as-is
- `examples`: `jupyterlab` + `reanalysis` + `nrel-wind`, which would essentially act as your `all-deps`
- `test`: delete; was required for maintaining a now-deprecated `setup.py` setting (learned this recently when porting `setup.py` to `pyproject.toml` for another project)
- `docs`: all documentation dependencies
- `develop`: current linting and testing deps + `docs` + `examples`

My thinking on keeping `nrel-wind` and `reanalysis` is that they're used in specific sections of the code, and don't inherently need to be wrapped up exclusively in the examples modifier, but are needed for running the examples, so adding them to `examples` satisfies keeping a minimum dependency stack (e.g. avoiding jupyter unless necessary) and having the separation users might need for their use cases.

jordanperr:
This plan sounds good to me. I'm fine keeping in any and all of the dependency modifiers, as long as new users have a clear pathway to install what they need and aren't confused or having to run pip install multiple times. As for Python 3.11 in the docs, that was my mistake based on misreading setup.py. Go ahead and roll that back.

PR #265, opened 2023-12-22
Include warnings about lack of validation of static yaw misalignment method

ejsimley:
This pull request clarifies that the `StaticYawMisalignment` method has not been validated on turbines with known static yaw misalignments, explains that the results should be treated with caution, and points out a potential issue with the method. The following language has been added to the `StaticYawMisalignment` docstring and `07_static_yaw_misalignment.ipynb` notebook. Additionally, a brief warning about the lack of validation has been added to the documentation homepage. 

"The Static Yaw Misalignment method in OpenOA is a relatively simple method that has not yet been validated using data from wind turbines with known static yaw misalignments. Therefore, the results should be treated with caution. One known issue is that the method currently relies on nacelle wind speed measurements to determine the power performance as a function of wind vane angle. If the measured wind speed is affected by the amount of yaw misalignment, potential biases can exist in the estimated static yaw misalignment values." 

For example, using the original data from one of the turbines from the example La Haute Borne data set, the static yaw misalignment is estimated as only 0.1 degrees for the 6 m/s wind speed bin:
image

Now, let's assume that the measured nacelle wind speed is affected asymmetrically by the amount of yaw misalignment by multiplying the wind speed by the following function of the wind vane angle:  
image

The resulting estimated static yaw misalignment is now -3.9 degrees, illustrating how yaw misalignment-dependent wind speed measurement biases can cause an overall bias in the estimated static yaw misalignment. Specifically, the wind speed measurements are lower than the true wind speed during negative yaw misalignment and greater than the true wind speed during positive yaw misalignment. Consequently, wind turbine power production values that are greater than expected based on the measured wind speed are included in negative yaw misalignment bins, shifting the peak of the power performance curve toward more negative wind vane angles.
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 All notable changes to this project will be documented in this file. If you make a notable change to the project, please add a line describing the change to the "unreleased" section. The maintainers will make an effort to keep the [Github Releases](https://github.com/NREL/OpenOA/releases) page up to date with this changelog. The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## [3.0.1 - 2023-12-22]
+- Includes warnings about limitations and lack of validation of static yaw misalignment method.
+
 ## v3.0 - 29 September 2023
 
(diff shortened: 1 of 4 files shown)


RHammond2 (approved):
This is great, @ejsimley! Let's get this merged and tag a patch release on this, so it's now 3.0.1.

In `__init__.py`, line 1, just update it to `__version__ = "3.0.1"`, then this can be merged into main.

Issue #264, opened 2023-12-09
Some cells in the example notebooks take a long time to run.


jordanperr:
I noticed the following cells take a relatively long time to run for an example:

**02c_augmented_plant_aep_analysis.ipynb:**
- `aep_gam.run(num_sim=1000)` takes about 5 minutes to run on my machine.

**03_turbine_ideal_energy.ipynb:**
- `ta.run(reanalysis_products=['era5', 'merra2'])` towards the end of the notebook takes 3 minutes to run on my machine.

**05_eya_gap_analysis.ipynb:**
- Cell that starts with `# Calculate TIE` also takes a long time (3 minutes) to run on my machine.

RHammond2:
@jordanperr this is how the examples have always been, except in v2 they took 15-20 minutes to run, so it'd be great to get more background on what exactly a satisfactory resolution would be. The primary reason they take a long time to run is to demonstrate the UQ capabilities of the models

jordanperr:
This has been an ongoing back-and-forth with these tests and examples. We have discussed the pros and cons in  reducing the number of iterations to speed up tests and examples but also wanting to maintain validity. Five minutes isn't too bad. Maybe a middle-ground is to add some text in the preceding notebook cell explaining that compute time can be reduced with the num_sim parameter (for a reduction in accuracy).

RHammond2:
@jordanperr, please check out #266 to see if this appropriately addresses the problem. I was aiming to balance run time and results stability based on the resulting plots, but let me know if anything should be changed further there.

PR #266, opened 2023-12-23
Enhancement: Reduce Example Runtimes

RHammond2:
This PR is based on #264 and does two things:
- Reduces the number of simulations performed in
  - `02c_augmented_plant_aep_analysis.ipynb`
  - `03_turbine_ideal_energy.ipynb`
  - `05_eya_gap_analysis.ipynb`

As was discussed in a meeting, Jupyter warnings and notes look awkward, so this will also address the formatting issues by adopting the MyST plugin for Jupyter Lab. The addition enables admonitions to look like they would in the documentation itself, which relies on MyST.

This will move out of a draft PR once #265 is merged in, so the formatting can be updated there as well.


ejsimley (approved):
Looks good to me!

Issue #235, opened 2023-05-08
Is there same defenition for oa.tie & eya.tie?


hamiddatalev:
The issue is:  "eya.tie" is calculated from "eya.gross energy" excluding some losses(line 125, eya_gap_analysis.py) but "oa.tie" is calculated from MC simulation at turbine level energy with no losses deducted from it. 
So, are these two KPIs comparable?

ejsimley:
Hi @hamiddatalev, the EYA and operational TIE values are intended to be comparable. Turbine Ideal Energy is meant to represent the energy the wind plant would be able to produce if all turbines were operating normally (i.e., always available and not curtailed or derated) and there were no major environmental losses. We define this as the gross energy of the wind plant with wake losses, turbine performance losses, and blade degradation losses subtracted. This definition is exactly how it's calculated from the EYA data because we have access to the predicted gross energy and loss categories. But in the operational data the losses are inherent in the turbine-level SCADA data, which is why we don't subtract any additional losses from the measurements. Instead, we use power curve filtering to identify times when the turbines are operating normally (not derated, curtailed, or underperforming significantly), then use those times to estimate the long-term ideal energy that the plant would be able to produce.

The method and a comparison between the EYA and OA TIE values is provided in the following journal article (see Section 3.1 for a detailed description of the methods for estimating TIE from EYA and operational data): https://onlinelibrary.wiley.com/doi/10.1002/we.2768

hamiddatalev:
•	Is the power curve filtering running in the "filter_turbine_data" function? and if yes:
o	"turb_capac" (line290, turbine_long_term_gross_energy.py) is calculated from max power, and I couldn't find any outlier detection filter used in "turbine_long_term_gross_energy.py" or "porject_ENGIE.py" for SCADA data. So, how do you deal if this max power is an outlier in the energy data of each turbine?
 because all other filters are based on this "turb_capac". 

•	How does this tool work for a site with multiple turbine OEMs or multiple rated powers?

ejsimley:
Yes, power curve outlier detection is performed in the `filter_turbine_data` function, mostly using the `bin_filter` function. That's a good point that the `turb_capac` parameter is based on the maximum power, so it is assuming the maximum power is not an outlier. I think this could be fixed by setting `turb_capac` equal to the "rated_power" value for each turbine from the asset table, so we'll review that choice and probably make that fix.

[...]

PR #268, opened 2024-02-02
Enhancement/update rated power in tie

ejsimley:
This pull request addresses the concern raised in issue #235 for the `TurbineLongTermGrossEnergy` analysis method. When filtering the SCADA data to identify periods with normal operation, the power curve outlier filter was previously using the maximum power from SCADA for each turbine to define the turbine capacity. This could create an issue in the `filters.bin_filter` function call if there happen to be any power outliers in the SCADA data significantly greater than the turbine's rated power. To fix this, the turbine capacity is changed to the rated power for each turbine given by the asset table.

Additional changes include:

- Removing a leftover, unused method `prepare_scada` in `TurbineLongTermGrossEnergy` that was replaced by `filter_turbine_data`
- Updating the TIE test results, because the minor change in the turbine capacity parameter causes the TIE results to change slightly 
- Re-running the example 03 notebook because the results changed slightly
- Updating the analysis requirements for `TurbineLongTermGrossEnergy` to include the asset table in metadata.py


RHammond2 (approved):
Thanks for being on top of this! Let's merge this in after #271, which will fix the test failures in the CI.

After merging develop back in after #271 is completed, just rerun the schema file: `python openoa/schema/schema.py` to fix any bad conflict resolutions in the schema YAML and JSON files.

PR #271, opened 2024-02-06
Bug Fix: Update Pandas Offsets Usage

### LINKED ISSUE #269: Update Pandas Offset Strings
While running test on the main branch in a fresh environment, 2 deprecation warnings have been raised or Pandas offset strings: 1) "T" should be replaced with "min", and 2) "H" should be replaced with "h".

```python
test/unit/test_filter_toolkit.py::SimpleFilters::test_bin_filter
  <path-removed>/envs/openoa-test/lib/python3.9/site-packages/numpy/lib/nanfunctions.py:1879: RuntimeWarning: Degrees of freedom <= 0 for slice.
    var = nanvar(a, axis=axis, dtype=dtype, out=out, ddof=ddof,

test/unit/test_plant_helpers.py::test_dtype_converter
   <path-removed>/OpenOA/test/unit/test_plant_helpers.py:191: FutureWarning: 'T' is deprecated and will be removed in a future version, please use 'min' instead.
    df.time = pd.date_range(start="2022-July-25 00:00:00", end="2022-July-25 1:00:00", freq="10T")

test/unit/test_timeseries_toolkit.py::SimpleTimeseriesTests::test_gap_fill_data_frame
test/unit/test_timeseries_toolkit.py::SimpleTimeseriesTests::test_gap_fill_data_frame
   <path-removed>/OpenOA/openoa/utils/timeseries.py:213: FutureWarning: The behavior of DataFrame concatenation with empty or all-NA entries is deprecated. In a future version, this will no longer exclude empty or all-NA columns when determining the result dtypes. To retain the old behavior, exclude the relevant entries before the concat operation.
    return pd.concat([data, gap_df], axis=0).sort_values(dt_col)

test/unit/test_timeseries_toolkit.py::SimpleTimeseriesTests::test_num_hours
test/unit/test_timeseries_toolkit.py::SimpleTimeseriesTests::test_num_hours
test/unit/test_timeseries_toolkit.py::SimpleTimeseriesTests::test_num_hours
   <path-removed>/OpenOA/openoa/utils/timeseries.py:264: FutureWarning: 'H' is deprecated and will be removed in a future version, please use 'h' instead.
    return dt_col[~dt_col.index.duplicated()].resample("H").asfreq().index.size

test/unit/test_unit_conversion_toolkit.py::SimpleUnitConversionTests::test_convert_power_to_energy
   <path-removed>/OpenOA/openoa/utils/unit_conversion.py:32: FutureWarning: 'T' is deprecated and will be removed in a future version, please use 'min' instead.
    _dt_range = pd.date_range(start="09/30/2022", periods=2, freq=sample_rate_min)

test/unit/test_unit_conversion_toolkit.py::SimpleUnitConversionTests::test_convert_power_to_energy
   <path-removed>/OpenOA/openoa/utils/unit_conversion.py:32: FutureWarning: 'H' is deprecated and will be removed in a future version, please use 'h' instead.
    _dt_range = pd.date_range(start="09/30/2022", periods=2, freq=sample_rate_min)
```

RHammond2:
This PR addresses the underlying issue causing all CI tests to fail in current PRs. The issue stems from a change in expected datetime offsets used in Pandas. From #269:

 - N -> ns
 - S -> s
 - T -> min
 - H -> h
 - U -> us
 - L -> ms

All underlying code that relied on the previously acceptable codes have been replaced. This will inevitably lead to a minor version bump due to some changes in usage.

Closes #269


ejsimley (approved):
Thanks for finding this change in the datetime offsets! This all looks good to me.  

Just wanted to note that we made some updates to the 07_static_yaw_misalignment.ipynb notebook in the main branch that aren't in develop yet. Since the example notebooks are changed in this PR too, we should make sure all relevant updates to the notebook are kept when we release the next version.

RHammond2:
Thanks for pointing that out, @ejsimley, I'll merge main back into develop, and update this PR again.

jordanperr:
Upgrading to new-style pandas time offsets will break the usage of old-style pandas time offsets that are passed through in OpenOA's API, or used outside of the API in user's code. For example, the "freq" argument in utils.qa.daylight_savings_plots, the "offset" argument in utils.timeseries.offset_to_seconds, and possibly the "time_resolution" argument in MonteCarloAEP.

To remain consistent with our semantic versioning, we should provide the users a fall-back to the old-style offsets if needed. One way this could work is to add a note about this deprecation issue to the Readme, advising users to downgrade Pandas themselves, and then add custom support for the old-style frequencies in our API (could simply translate the old strings to the new ones).

Or, alternatively, just pin Pandas below the version where this breaks for the 3.x line, silence the warning, and put this PR in the 4.x release.

RHammond2:
@jordanperr I was able to add in some support for the "old" style of offset strings without much extra lift. The catch is that it will output a deprecation warning pushing users to adopt the "new" style. I also added a test for the method that is applied to the `MonteCarloAEP.time_resolution` attribute and the `frequency` attribute for the metadata classes.

PR #272, opened 2024-02-06
Bug Fix: Pandas loc-chaining

### LINKED ISSUE #270: Fix Project ENGIE Cleaning and Pandas Nested `.loc` Warning
There are two overlapping issues that can't be disentangled, and should be addressed simultaneously:
1) Pandas will eventually stop supporting nested indexing as it cleans up its backend.
2) In the v3 cleanup, a reassignment in a nested `.loc` was improperly removed see [`project_ENGIE.py:98`](https://github.com/NREL/OpenOA/commit/89f3dbe61747ad0a0075887836cb4599ffcc9064#diff-dd51ca949ed033918087905669f2893fd51fc77b272bead5a72ad2b8716dedcfL98).

Both should be rewritten to the form of, but in 2) the ` = np.nan` is missing:

```python
scada_df.loc[ix_flag.loc[ix_flag["<focal_column>"]].index, sensor_cols] = np.nan
```

The change in form won't the results, however, because in line 99 of the current file the assignment part of the expression is missing, our results will naturally change, and as a result, may be where some of our change in results earlier on were stemming from.

RHammond2:
This PR resolves #270 by combining the chained `.loc` statements into a single `.loc` and addresses the missing NaN assignment in `project_ENGIE.py:clean_scada()`.

An additional deprecation warning is also addressed by not concatenating empty data frames in the timeseries gap filling methods.


ejsimley:
Thanks for fixing this @RHammond2!

PR #273, opened 2024-02-06
Maintenance: Complete Adoption of `project.toml`

RHammond2:
This PR converts `setup.py` to `pyproject.toml` for package management and fixes a minor bug in the `pytest --unit` and `pytest --regression` custom options.

PR #274, opened 2024-02-06
Maintenance: Upgrade Outdated Python Style

RHammond2:
This PR is the result of running `pyupgrade --py38-plus` against the codebase to identify lingering pre-3.8 language standards.

TODO prior to moving out of draft stage:
- [x] Incorporate develop following at least PRs #271, #272, #273, and #266
- [ ] Run the plotting examples to ensure the updated f-string format annotations still work following the pyupgrade styling


openoa/plant.py
```diff
@@ -148,7 +148,7 @@ def _compose_error_message(
 @logged_method_call
 def frequency_validator(
     actual_freq: str | int | float | None,
-    desired_freq: Optional[str | None | set[str]],
+    desired_freq: str | None | set[str] | None,
```
ejsimley, line 151: None appears twice. Is this a typo?

RHammond2: Good catch! I must've missed that one when reviewing.

ejsimley (approved):

PR #282, opened 2024-03-07
Bug Fix: Asset table operations

ejsimley:
This small pull request fixes a few issues I found when calling the `get_freestream_turbines` method in `PlantData` and running wake loss analyses for a wind plant with both turbine and tower assets.

1. In `calculate_asset_distance_matrix` and `calculate_asset_direction_matrix`, the diagonals of the matrices weren't getting set to np.inf like we expected. This could possibly be because of a new pandas version. 
2. An error would appear in `get_freestream_turbines` when there are tower assets in addition to turbine assets. Specifically, `self.asset.index` includes both turbine and tower indices, but `freestream_indices` only includes boolean values corresponding to turbines, so the dimensions didn't match in the return statement.

### DIFF
--- a/openoa/plant.py
+++ b/openoa/plant.py
@@ -1252,7 +1252,9 @@ class PlantData:
 
         # Maintain v2 compatibility of np.inf for the diagonal
         distance = distance + distance.values.T - np.diag(np.diag(distance.values))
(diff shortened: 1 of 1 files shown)


RHammond2 (approved):
Thanks for catching this, @ejsimley! This fix looks good to me.

PR #296, opened 2024-06-04
Hourly reanalysis downloading

ejsimley:
This pull request adds two new functions to the `utils/downloader` module to download hourly reanalysis data: `get_era5_hourly` and `get_merra2_hourly`. These functions are modified from `get_era5_monthly` and `get_merra2_monthly` contributed by @charlie9578, and similarly download data from the CDS and GES DISC services. 

Note that it can take a long time to download historical data (~1 day for a 20-year time series). Downloading era5 data seems to be faster than merra2 for me, though.


openoa/utils/downloader.py
```diff
@@ -333,6 +336,248 @@ def get_era5_monthly(
...
+        Exception: If unable to connect to the cdsapi client.
+    """
+
+    logger.info("Please note access to ERA5 data requires registration")
+    logger.info("Please see: https://cds.climate.copernicus.eu/api-how-to")
+
+    # set up cds-api client
+    try:
+        c = cdsapi.Client()
+    except Exception as e:
```
RHammond2, line 397: Is this what's raised from the API? I thought it was more specific for some reason, or are you just trying to capture any failed connection, regardless of reason?

openoa/utils/downloader.py
```diff
@@ -333,6 +336,248 @@ def get_era5_monthly(
...
+        "format": "netcdf",
+        "variable": [
+            "100m_u_component_of_wind",
+            "100m_v_component_of_wind",
+            "2m_temperature",
+            "surface_pressure",
+        ],
+        "year": None,
+        "month": None,
+        "day": [
```
RHammond2, line 446: I wonder if we could do `[f"{i:02d}" for i in range(1, 32)]` just to trim the size of the list

openoa/utils/downloader.py
```diff
@@ -333,6 +336,248 @@ def get_era5_monthly(
...
+            "25",
+            "26",
+            "27",
+            "28",
+            "29",
+            "30",
+            "31",
+        ],
+        "time": [
+            "00:00",
```
RHammond2, line 480: Same here, but with `[f"{i:02d}:00" for i in range(24)]`.

openoa/utils/downloader.py
```diff
@@ -333,6 +336,248 @@ def get_era5_monthly(
...
+
+    This function returns hourly ERA5 data from the "ERA5 monthly averaged data on single levels
+    from 1959 to present" dataset. See further details regarding the dataset at:
+    https://cds.climate.copernicus.eu/cdsapp#!/dataset/reanalysis-era5-single-levels-monthly-means.
+    U and V components of wind speed at 100 m, temperature at 2 m, and surface pressure are
+    downloaded here.
+
+    As well as returning the data as a dataframe, the data is also saved as monthly NetCDF files and
+    a csv file with the concatenated data. These are located in the "save_pathname" directory, with
+    "save_filename" prefix. This allows future loading without download from the CDS service.
```
RHammond2, line 360: I was thinking about this, and I think we should just delete the monthly files, and save as one single netcdf and csv file as it's less onerous on the user who'll have to delete the originals if they're no longer required.

openoa/utils/downloader.py
```diff
@@ -482,3 +727,182 @@ def get_merra2_monthly(
...
+            else:
+                months = list(range(start_date.month, 12 + 1, 1))
+        elif year == end_date.year:
+            months = list(range(1, end_date.month + 1, 1))
+        else:
+            months = list(range(1, 12 + 1, 1))
+
+        for month in months:
+            # get the file names from the GES DISC site for the year
+            result = requests.get(base_url + str(year) + "/%02d" % month)
```
RHammond2, line 828: The string portion could get tidied up to` f"{base_url}{year}/{month:02d}"`
 if I followed that correctly.

openoa/utils/downloader.py
```diff
@@ -482,3 +727,182 @@ def get_merra2_monthly(
...
+        # limit to months of interest
+        if year == start_date.year:
+            if year == end_date.year:
+                months = list(range(start_date.month, end_date.month + 1, 1))
+            else:
+                months = list(range(start_date.month, 12 + 1, 1))
+        elif year == end_date.year:
+            months = list(range(1, end_date.month + 1, 1))
+        else:
+            months = list(range(1, 12 + 1, 1))
```
RHammond2, line 824: The extra 1 at the end is the default, even when the starting point is changed. Though if that's just a style thing, then it doesn't bother me, but thought I'd flag it just in case.

RHammond2 (commented):
Thanks for putting this together, it's a great addition to the toolkit! The only comment holding me back from an approval was the one regarding the monthly files from ERA5, and then all the base files from the MERRA2. I'm thinking we don't really need them, and should ditch them, but I could be swayed towards leaving it as-is as well.

openoa/utils/downloader.py
ejsimley: This was just copied from the existing monthly downloading function. I think the point was to print a link where you can register for the CDS API when unable to connect. I decided to change this to the try/except statement for certificate verification that you showed me. If unable to connect, the error message from the CDS API should at least still provide some useful information.

openoa/utils/downloader.py
```diff
@@ -333,6 +336,248 @@ def get_era5_monthly(
...
+            "21:00",
+            "22:00",
+            "23:00",
+        ],
+        "product_type": "reanalysis",
+        "area": [
+            lat + node_spacing,
+            lon - node_spacing,
+            lat - node_spacing,
+            lon + node_spacing,
```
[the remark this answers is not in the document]
ejsimley: The original intent was to download data for an area of 2\*node_spacing x 2\*node_spacing around the specified coordinates so that the nearest grid point (which might not be known beforehand) can be selected form the 9 grid points that are returned. However, I found that the ERA5 API request for an arbitrary lat/lon will linearly interpolate between the nearest grid points, so this didn't return data from the intended grid point. I replaced this with code to return data from the nearest point on the 0.25 deg. by 0.25 deg. grid. Since it is no longer necessary to download data for an area around the specified grid point, the request only asks for data at the nearest grid point now.

ejsimley:
@RHammond2, thanks for your ideas for improving this code. I think I addressed all your comments, so this is ready for a re-review. In addition to the code style improvements you pointed out, here are some of the main changes I made:

- The API request and documentation for ERA5 has been updated to reflect the recent CDS API changes.
- The API requests for both MERRA2 and ERA5 now only get data from the nearest grid point to the specified coordinates. Since I found that the ERA5 data that gets returned for an arbitrary coordinate is linearly interpolated from the nearest grid points (which is not what we intended when first writing the functions), I updated the code to explcitly request data for the nearest grid point.
- The intermediate NetCDF files that get downloaded are now deleted after the csv files get saved.
- The above changes were also made to the existing monthly reanalysis downloading functions.

openoa/utils/downloader.py
```diff
@@ -333,6 +336,248 @@ def get_era5_monthly(
...
+    # check that the start and end dates are the right way around
+    if start_date > end_date:
+        logger.error("The start_date should be less than or equal to the end_date")
+        logger.error(f"start_date = {start_date.date()}, end_date = {end_date.date()}")
+        raise ValueError("The start_date should be less than or equal to the end_date")
+
+    # list all years that will be downloaded
+    years = list(range(start_date.year, end_date.year + 1, 1))
+
+    # get the data for the closest 9 nodes to the coordinates
```
RHammond2: This comment needs to be updated.

openoa/utils/downloader.py
```diff
@@ -258,10 +265,13 @@ def get_era5_monthly(
...
 
-    # get the data for the closest 9 nodes to the coordinates
-    node_spacing = 0.250500001 * 1
+    # find the nearest coordinate grid point
+    node_spacing = 0.25
+    lat_nearest = node_spacing * np.round(lat / node_spacing)
+    lon_nearest = node_spacing * np.round(lon / node_spacing)
 
-    # See: https://cds.climate.copernicus.eu/cdsapp#!/dataset/reanalysis-era5-single-levels-monthly-means?tab=form
+    # Get data for 9 nearest grid points
```
RHammond2, line 273: This comment can get removed.

openoa/utils/downloader.py
```diff
@@ -330,6 +340,200 @@ def get_era5_monthly(
...
+    try:
+        c = cdsapi.Client()
+    except SSLError:
+        print("Skipping certificate verification")
+        c = cdsapi.Client(verify=False)  # verification error for self-signed certificate
+
+    # create save_pathname if it does not exist
+    save_pathname = Path(save_pathname).resolve()
+    if not save_pathname.exists():
+        save_pathname.mkdir()
```
RHammond2, line 416: Do we need `parents=True` to create any missing directories higher up than the last folder?

openoa/utils/downloader.py
```diff
@@ -330,6 +340,200 @@ def get_era5_monthly(
...
+            "2m_temperature",
+            "surface_pressure",
+        ],
+        "year": None,
+        "month": None,
+        "day": [f"{i:02d}" for i in range(1, 32)],
+        "time": [f"{i:02d}:00" for i in range(24)],
+        "product_type": "reanalysis",
+        "area": [
+            lat_nearest + 0.01,
```
RHammond2, line 463: What is the 0.1 buffer for? Unless something has changed, I believe a single point can be requested by repeating the latitude and longitude.

openoa/utils/downloader.py
```diff
@@ -330,6 +340,200 @@ def get_era5_monthly(
...
+
+            try:
+                cds_request.update({"year": year, "month": months})
+                c.retrieve(cds_dataset, cds_request, outfile)
+            except Exception as e:
+                logger.error(f"Failed to download ERA5: {outfile}")
+                logger.error(e)
+
+    # get the saved data
+    ds_nc = xr.open_mfdataset(f"{save_pathname / f'{save_filename}*.nc'}")
```
RHammond2, line 496: This works, but just a note that xarray, pandas, and pretty much any major package will handle `Path` object file inputs at this point.

RHammond2 (approved):
A couple of really minor comments, otherwise thanks for cleaning this up @ejsimley! I think this is good to go whenever you're comfortable with it.

charlie9578:
Don't hate me... but I just found this was released last week: [era5-timeseries](https://cds.climate.copernicus.eu/datasets/reanalysis-era5-single-levels-timeseries?tab=download), which significantly speeds up era5 downloads... although at the moment the signals are more limited and its in a beta stage... so perhaps something to consider later on!

Note too that some consultants/analysts use the 4/9 nearest nodes, because sometimes the nearest node doesn't correlate as well as one further away. For example with a costal wind farm where the node offshore/onshore correlates better than the one nearest to the wind farm in a different environment. I never applied this in the OpenOA code though, but was a legacy from some internal code where I did.

charlie9578:
Don't hate me... but I just found this was released last week: [era5-timeseries](https://cds.climate.copernicus.eu/datasets/reanalysis-era5-single-levels-timeseries?tab=download), which significantly speeds up era5 downloads (from hours to <30s), and in (zipped) csv format too... although at the moment the signals are more limited and its in a beta stage... so perhaps something to consider later on!

Note too that some consultants/analysts use the 4/9 nearest nodes, because sometimes the nearest node doesn't correlate as well as one further away. For example with a costal wind farm where the node offshore/onshore correlates better than the one nearest to the wind farm in a different environment. I never applied this in the OpenOA code though, but was a legacy from some internal code where I did.

RHammond2:
@charlie9578 that's amazing news for ERA5! It seems like the functionality and data may change over time, so it's something we should keep watching for when it becomes stable. The download times alone are a huge selling point, though It looks like there is only 10m u/v wind, and not 100m so we wouldn't be able to extrapolate wind speeds.

ejsimley:
Hi @charlie9578, thanks also for explaining why the 9 nearest nodes were getting requested in the monthly download functions. I saw that only the nearest node was getting used, so simplified the request in the hourly version to only get the nearest node. We could consider downloading the 4/9 nearest nodes as an option in the future though.

charlie9578:
> @charlie9578 that's amazing news for ERA5! It seems like the functionality and data may change over time, so it's something we should keep watching for when it becomes stable. The download times alone are a huge selling point, though It looks like there is only 10m u/v wind, and not 100m so we wouldn't be able to extrapolate wind speeds.

Also, for reference, and if you want to add your voice, there's a forum post regarding the time-series [here ](https://forum.ecmwf.int/t/new-dataset-published-in-cds-era5-hourly-time-series-data-on-single-levels-from-1940-to-present/11919?u=charlie_plumley). Thanks, Charlie

openoa/utils/downloader.py
ejsimley: It looks like that array function will only accept glob-like file paths if they are strings. I get an error when leaving it as a `Path` object. I cleaned it up a little by just wrapping `str()` around the `Path` object.

openoa/utils/downloader.py
ejsimley: You're right, the buffer doesn't seem to be needed.

ejsimley:
@RHammond2, thanks for taking another look. I tried to address your remaining comments as best I could. Note that I also changed the ERA5 hourly downloading function so that it requests data in month-long chunks from the CDS API instead of year-long requests. I started getting error messages saying the requests were too large, but found that 1-month (or up to around 4 months) was fine. This doesn't affect performance much, so I'm going to go ahead and merge.

PR #307, opened 2025-04-24
Enhancement: Add flexibility to the Monte Carlo analysis PlantData initialization

RHammond2:
This PR addresses a series of small features and fixes from the HVS2 work as outlined below.

- `MonteCarloAEP`
  - `n_jobs` to enable parallelization of each iteration's model
  - `apply_iav` to toggle the addition of the interannual variability at the end of the analysis.
  - `run::progress_bar` to toggle the use of a progress bar during the analysis. When running many models, turning off the progress bar for each iteration of a parameter sweep greatly reduces the amount of outputs to scroll through.
- Implements and applies a wind speed calculation from the u and v components in place of a manual calculation.
- Adds default empty data frames for the `PlantData`'s `asset_distance_matrix` and `asset_direction_matrix` values. Without the defaults, plants not utilizing location data end up failing at analysis time since no data is able to be passed.

### DIFF
--- a/openoa/analysis/aep.py
+++ b/openoa/analysis/aep.py
@@ -120,6 +120,9 @@ class MonteCarloAEP(FromDictMixin, ResetValuesMixin):
             points. Defaults to "lin".
         ml_setup_kwargs(:obj:`kwargs`): Keyword arguments to
             :py:class:`openoa.utils.machine_learning_setup.MachineLearningSetup` class. Defaults to {}.
+        n_jobs(:obj:`int` | :obj:`None`): The number of jobs to use for the computation in the scikit-learn model.
+            This will only provide speedup in case of sufficiently large problems.``None`` means 1
+            unless in a :obj:`joblib.parallel_backend` context. ``-1`` means using all processors.
     """
 
     plant: PlantData = field(converter=deepcopy, validator=attrs.validators.instance_of(PlantData))
@@ -174,6 +177,10 @@ class MonteCarloAEP(FromDictMixin, ResetValuesMixin):
         default="lin", converter=str, validator=attrs.validators.in_(("lin", "gbm", "etr", "gam"))
     )
     ml_setup_kwargs: dict = field(default={}, converter=dict)
+    n_jobs: int | None = field(
+        default=None, validator=attrs.validators.instance_of((int, type(None)))
+    )
+    apply_iav: bool = field(default=True, validator=attrs.validators.instance_of(bool))
 
     # Internally created attributes need to be given a type before usage
     resample_freq: str = field(init=False)
@@ -929,7 +936,9 @@ class MonteCarloAEP(FromDictMixin, ResetValuesMixin):
         # Run regression. Note, the last column of reg_data is the target variable for the regression
         # Linear regression
         if self.reg_model == "lin":
-            reg = LinearRegression().fit(np.array(reg_data[:, 0:-1]), reg_data[:, -1])
+            reg = LinearRegression(n_jobs=self.n_jobs).fit(
+                np.array(reg_data[:, 0:-1]), reg_data[:, -1]
+            )
             predicted_y = reg.predict(np.array(reg_data[:, 0:-1]))
 
             self._mc_slope[n, :] = reg.coef_
@@ -958,6 +967,7 @@ class MonteCarloAEP(FromDictMixin, ResetValuesMixin):
                     report=False,
                     cv=KFold(n_splits=5),
                     verbose=verbosity,
+                    n_jobs=self.n_jobs,
                 )
                 # Store optimized hyperparameters for each reanalysis product
                 self.opt_model[(self._run.reanalysis_product)] = ml.opt_model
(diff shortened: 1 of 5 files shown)


openoa/analysis/aep.py
```diff
@@ -324,6 +337,8 @@ def run(
                 points. Defaults to "lin".
             ml_setup_kwargs(:obj:`kwargs`): Keyword arguments to
                 :py:class:`openoa.utils.machine_learning_setup.MachineLearningSetup` class. Defaults to {}.
+            progress_bar(:obj:`bool`): Flag to use a progress bar for the iterations in the AEP
+                calculation.
```
ejsimley, line 341: Add "Defaults to True."

openoa/analysis/aep.py
```diff
@@ -959,10 +981,14 @@ def run_regression(self, n):
...
 
     @logged_method_call
-    def run_AEP_monte_carlo(self):
+    def run_AEP_monte_carlo(self, progress_bar: bool = True):
         """
         Loop through OA process a number of times and return array of AEP results each time
 
+        Args:
+            progress_bar(:obj:`bool`): Flag to use a progress bar for the iterations in the AEP
+                calculation.
```
ejsimley, line 990: Add "Defaults to True."

openoa/utils/machine_learning_setup.py
```diff
@@ -107,6 +107,9 @@ class MachineLearningSetup:
     random_search: Any = field(init=False)
     opt_hyp: Any = field(init=False)
     opt_model: Any = field(init=False)
+    n_jobs: int | None = field(
```
ejsimley, line 110: Since n_jobs is only used as an argument in the `hyper_optimize` method, is there a reason to list it as a class attribute here? If so, should this be described in the docstring, indicating that it can be specified when initializing an object of the class.

openoa/utils/met_data_processing.py
```diff
@@ -70,6 +70,27 @@ def circular_mean(x: pd.DataFrame | pd.Series | np.ndarray, axis: int = 0):
     )
 
 
+@series_method(data_cols=["u", "v"])
+def compute_wind_speed(
+    u: pd.Series | str, v: pd.Series | str, data: pd.DataFrame = None
+) -> pd.Series:
+    """Compute the wind speed from the u and v components. This function should only be used for
+    higher resolution wind data, otherwise the resulting wind speed will be lower than expected.
```
ejsimley, line 78: I'd suggest a little explanation about the problem with using lower temporal resolution data to help users decide what resolution is acceptable. For example, vector-averaged u and v wind speed components will generally result in lower wind speed magnitudes than averages of the wind speed magnitude over the same time period. This becomes more severe as the averaging period increases.

openoa/utils/met_data_processing.py
```diff
@@ -70,6 +70,27 @@ def circular_mean(x: pd.DataFrame | pd.Series | np.ndarray, axis: int = 0):
...
+
+    Args:
+    u(:obj:`pandas.Series` | :obj:`str`): A pandas DataFrame or Series, or a numpy array
```
ejsimley, line 81: For the "u" and "v" arguments, if a numpy array is allowed, can this be added to the function header and the list of argument types in the docstring too?

Can you also describe the string argument type here, like in the `compute_wind_direction` argument? ("or the name of the column in :py:attr:`data`")

openoa/utils/met_data_processing.py
```diff
@@ -70,6 +70,27 @@ def circular_mean(x: pd.DataFrame | pd.Series | np.ndarray, axis: int = 0):
...
+        containing the u-component of the wind speed, in m/s.
+    v(pd.Series | str): A pandas DataFrame or Series, or a numpy array containing the v-component
+        of the wind speed, in m/s.
+    data(:obj:`pandas.Series` | :obj:`str`): The pandas ``DataFrame`` containg the columns
+        :py:attr:`u` and :py:attr:`v`.
+
+    Returns:
+        :obj:`pandas.Series`: wind speed, in m/s.
```
ejsimley, line 89: Should this list `numpy.array`as a possible return type too?

ejsimley (changes requested):
Thanks for submitting these fixes and improvements based on the HVS2 work! I appreciate the "apply_iav" option, which makes the code a lot more flexible and the new `compute_wind_speed` function.

I just have some minor comments to consider.

openoa/analysis/aep.py
RHammond2: Added.

openoa/utils/met_data_processing.py
RHammond2: Added your snippet as a note and removed the abbreviated description.

openoa/analysis/aep.py
RHammond2: Updated.

openoa/utils/met_data_processing.py
ejsimley: Looks good. I just added one more sentence. Feel free to change it though.

ejsimley (approved):
Thanks for making these changes @RHammond2! I just added a little more to a couple docstrings (feel free to revise).

PR #308, opened 2025-05-07
adding weto stack blurb to readme

gbarter:


RHammond2 (commented):
@gbarter should this update also be on the documentation site?

gbarter:
Thanks, @RHammond2 !  That is a good call.  I have now added the WETO Stack blurb to the docs too.

PR #309, opened 2025-05-29
Enhancement: Update supported Python versions

RHammond2:
This PR updates Python support from 3.8 through 3.11 to 3.9 through 3.12. Additionally, the note stemming from early 3.11 support has been removed as pip's resolver is much better than when the warning was written.

The only thing stopping OpenOA from supporting 3.13 is pyGAM, which has not had a new version release since February 2024. This library does not seem to have an immediate replacement, so further work must be done if pyGAM becomes deprecated.


ejsimley (approved):
Thanks for keeping the supported python versions up to date and checking the dependencies!

PR #322, opened 2026-02-14
Add Frontend and backend in codebase

govindjoshi9:


RHammond2:
Hi @govindjoshi9, thanks for submitting this PR. Would you be able to add a description for what this PR is trying to accomplish? As it stands this won't be merged without additional information for the purpose and scope of this work as we don't currently support any web-based services for running the models.

